PR #168, opened 2014-12-19
Prefer moment.fn.utcOffset to moment.fn.zone

timrwood:
Corresponding moment issue: https://github.com/moment/moment/pull/2074

### DIFF
--- a/tests/helpers/helpers.js
+++ b/tests/helpers/helpers.js
@@ -2,6 +2,14 @@
 
 var moment = require('../../index');
 
+function getUTCOffset (m) {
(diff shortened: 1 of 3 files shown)


EgoAleSum:
Next time, it would be nice to support the principles of semver, though. APIs should not change from 0.2.4 to 0.2.5.

timrwood:
@EgoAleSum, this was a deprecation from moment core. I believe `moment.fn.zone` will be supported until moment 3.0 according to semver.

This does not change the moment-timezone public api though, so semver doesn't really have anything to do with this PR.

PR #173, opened 2015-01-14
Removing countries in meta.json which don't have any timezones in them.

kashifshamaz21:
List of countries that gets added to meta data json files, is parsed from `iso3166.tab`.
And the mapping between which zones are present in which country is parsed from `zone1970.tab`

But some countries(Bouvet Island, Heard Island & McDonald Islands) there may not be any zones at all.
Those countries should be filtered out as it doesn't have any zones and are of no significance for timezones.


timrwood:
Looks great! I merged and rebased manually to squash those 4 commits into one here. a5bd3e177fbfaf39017887fb106ca195affd017f

PR #201, opened 2015-04-06
time zone data: 2015b

vmeurisse:


data/unpacked/latest.json
```diff
@@ -10537,15 +10534,15 @@
     {
       "name": "America/Cayman",
       "abbrs": [
-        "KMT",
+        "CMT",
```
mj1856: This isn't in the 2015b sources.  Where did it come from?

mj1856:
I see a few other discrepancies that aren't in 2015b.   I'm comparing against:
http://mm.icann.org/pipermail/tz-announce/2015-March/000029.html
https://github.com/eggert/tz/tree/2015b
http://www.iana.org/time-zones/repository/releases/tzdata2015b.tar.gz

@vmeurisse Can you please start clean and rebase your branch?  Thanks.

data/unpacked/latest.json
vmeurisse: This changed is caused by America/Cayman now being a link to America/Panama
https://github.com/eggert/tz/commit/72a3ac9cb075b934c13885b6ef2616374b704fe6#diff-9c77a7aa95323ac0135649098471287eL2600

vmeurisse:
@mj1856 I retried with a clean branch and both zdumb and zic compiled from https://github.com/eggert/tz/tree/2015b but I still get the same results

data/unpacked/latest.json
mj1856: @vmeurisse  - yep, looks like that was the reason.  Thanks.

mj1856:
Looks good.  Thanks for the explaination.

PR #211, opened 2015-05-06
unpack only what needed

davidpean:
Only execute the unpack for the one timezone if it hasn't been done already. Based on some testing, on initial load desktop is around ~2ms, and Chrome Android ~30 now

Previously the multiple unpacks were really expensive on mobile

screenshot 2015-05-06 12 28 04

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -24,6 +24,7 @@
 	var VERSION = "0.3.1",
 		zones = {},
 		links = {},
+		zonesByName = {},
 
 		momentVersion = moment.version.split('.'),
 		major = +momentVersion[0],
@@ -212,6 +213,9 @@
 	}
 
 	function getZone (name) {
+		if( zonesByName[name] && !zones[name] ){
+			addZone( zonesByName[name] );
+		}
 		return zones[normalizeName(name)] || null;
 	}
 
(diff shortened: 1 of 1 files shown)


paulirish:
The cost of moment-timezone loadData on ESPN.com

* 530ms before this patch
* 30ms after this patch

:+1:

mj1856:
I'll review this in more detail for possible side-effects, but on the surface it looks good.  Thanks!

timrwood:
This looks good, except for some weirdness with loading linked zones. If you use a linked zone before an actual zone, the data is not ready yet. http://jsfiddle.net/3tj066w7/ See how the output for `US/Central` changes once the `America/Chicago` timezone gets unpacked.

I think we could probably change [these lines](https://github.com/moment/moment-timezone/blob/532cf3e895268389998bd538a46d0c442d45adf9/moment-timezone.js#L269-L271) to something like this.

```js
var zone = getZone(zoneName)
if (zone) {
  copyZoneWithName(zone, linkName);
} else {
```

However, that would probably lose a lot of the perf benefits as each link would cause its related zone to be unpacked.

I have another approach at #215 that isn't quite as fast as this approach, however we should recheck the improvements once the out of order stuff is fixed.

davidpean:
@timrwood - Thanks Tim, I didn't realize what the links were for. Couldn't we just create a zoneByName of all the links? In loadData I created an entry since all the regular ones have already been mapped. This should allow the regular unpack to work regardless of source.

https://github.com/davidpean/moment-timezone/commit/4d152b50dee96c565a8353b044193f8c78aacf5e

Example: https://jsfiddle.net/rn58wmvn/

davidpean:
sorry, my jsfiddle had old source: https://jsfiddle.net/davidpean/rn58wmvn/2/

timrwood:
@davidpean, yeah, I ended up doing something similar over on #216.

I changed the implementation around a bit to pull when getting a zone rather than push when loading a zone. This was to support loading zones and links [in independent order of each other](https://github.com/moment/moment-timezone/blob/develop/tests/moment-timezone/link.js#L90-L109).

timrwood:
The version of this from #216 is released in `0.4.0`.

PR #205, opened 2015-04-16
time zone data: 2015c

mj1856:
tzdb 2015, release April 14, 2015
http://mm.icann.org/pipermail/tz-announce/2015-April/000030.html

PR #214, opened 2015-05-11
2015c and 2015d data updates

timrwood:
See #205


data/unpacked/latest.json
```diff
@@ -774,62 +718,6 @@
         1403816400000,
         1406844000000,
         1411678800000,
-        1429826400000,
```
mj1856: @timrwood - Why are the "untils" being truncated here?

mj1856:
I still see way more data updates in the unpacked files than are actually in the tzdb releases (as far as I can tell).   Are you just running `grunt data:2014c` or something different or more than that?

FYI - that's all I did, and was on a clean Ubuntu box when I did #205, so it's not a line-endings issue.  Currently we actually *can't* build the TZ files on Windows due to dependencies on `zic` and `zdump`.

Also - the [contributing guide](https://github.com/moment/moment-timezone/blob/develop/contributing.md) describes how to compile these files, but it doesn't describe how to build the `moment-timezone-2010-2020.js` and `moment-timezone-all-years.js` files to update the moment-timezone home page - which currently still references 2014j.  We need to get this all straightened out. (please)

data/unpacked/latest.json
timrwood: I also included 2015d which removed DST in Egypt. http://mm.icann.org/pipermail/tz-announce/2015-April/000031.html

mj1856:
Merging as is, though I could still use additional instruction in this area.  Thanks.

timrwood:
@mj1856, `moment-timezone-2010-2020.js` and `moment-timezone-all-years.js` are built with the `grunt release` command.

synotna:
Is there an ETA of these updates being released?

PR #215, opened 2015-05-11
Unpack only what needed (alternate approach)

timrwood:
Alternate approach to #211

Store the packed string and name on the Zone object.

When the zone is requested, either from getZone or pushLink, the data is unpacked if needed.
The packed property is then reset to ensure that data is not being re-unpacked.

This method ensures that the data is available whether the link or zone used first. See http://jsfiddle.net/3tj066w7/1/

This approach is not nearly as fast as the one in #211, but we should check again after fixing the load order issue.

Perf improvements for this approach. http://jsperf.com/moment-timezone-unpack-on-demand

Perf improvements for the #211 approach. http://jsperf.com/moment-timezone-unpack-on-demand-2

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -129,13 +129,21 @@
 
 	function Zone (packedString) {
 		if (packedString) {
-			this._set(unpack(packedString));
+			this.packed = packedString;
+			this.name = packedString.split('|')[0];
 		}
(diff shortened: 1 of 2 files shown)


timrwood:
Closing in favor of #216

Issue #212, opened 2015-05-07
Guard against loading moment-timezone twice


mj1856:
A few reports from the wild, such as in #193 and [on StackOverflow](http://stackoverflow.com/q/29877791/634824) (and others) have shown that it's easy to accidentally load moment-timezone twice, which creates some problems.

Usually, it's a misunderstanding that moment-timezone-with-data already contains moment-timezone.

We should either guard against this, or at least prevent it from causing problems.

timrwood:
This is released in `0.4.0`.

PR #219, opened 2015-05-12
Log an error if moment timezone was already loaded. #212

timrwood:
This adds an error message when loading moment-timezone twice. This addresses #212.

http://jsfiddle.net/eLh5m79c/

image

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -19,7 +19,10 @@
 	"use strict";
 
 	// Do not load moment-timezone a second time.
-	if (moment.tz !== undefined) { return moment; }
+	if (moment.tz !== undefined) {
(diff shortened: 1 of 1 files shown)


mj1856:
:+1:

jebeck:
Can someone clarify? I'm seeing this in the following situation. We have a (parent) application A. It has dependencies B and C, each of which has moment-timezone as a dependency. I'm getting this log message, and unsure whether I need to do anything about it. It looks like the code is preventing double-loading of the moment-timezone data, so I *think* everything is fine. If that's the case, shouldn't the error message be something more like "Moment Timezone [loaded version] was already loaded with data from [loaded version data]. Not loading [second version] with data [second version data]. Unless [first version] is not the version you wanted to be loaded, everything is fine."

Happy to issue a PR with this change (or similar) if I'm interpreting the situation correctly!

sunnybayz:
Agree with @jebeck . Should we worry about this error?

Issue #138, opened 2014-10-16
Get the browsers current time zone.


niemyjski:
Unless I'm missing something it seems that there is no way to get the current time zone?? I know there are third party libs to get the olson name.. I just thought this would be included.

   moment.tz().zoneName()
   moment.tz(new Date()).zoneName()
   moment.tz(new Date()).zone()

kashifshamaz21:
@elad Did you get a chance to wire up a working protoype for this feature.. I can give it a shot, if you haven't done a lot of progress on it already. Thanks!

elad:
@kashifshamaz21 please do! Unfortunately I haven't had a chance to get to it yet. :(

softBarbarian:
Hello, 
This feature would be awsome!
+1000

RouR:
Check this code
http://stackoverflow.com/questions/19420582/detect-the-id-of-the-current-user-timezone-using-moment-js
https://github.com/Canop/tzdetect.js

```javascript
var tzdetect = {
    names: moment.tz.names(),
    matches: function(base){
        var results = [], now = Date.now(), makekey = function(id){
            return [0, 4, 8, -5*12, 4-5*12, 8-5*12, 4-2*12, 8-2*12].map(function(months){
                var m = moment(now + months*30*24*60*60*1000);
                if (id) m.tz(id);
                return m.format("DDHHmm");
            }).join(' ');
        }, lockey = makekey(base);
        tzdetect.names.forEach(function(id){
            if (makekey(id)===lockey) results.push(id);
        });
        return results;
    }
};
```

If you just want one timezone id, simply use 
    var tzid = tzdetect.matches()[0];

mj1856:
I love the idea of having a function that can be used directly against the moment-timezone data, rather than jsTimeZoneDetect, which has it's own data.  However, I'm not convinced that the Canop/tzdetect solution is complete enough to be useful.

Running it on my computer, which is Windows 8.1, Chrome latest, in US Pacific time, I get:

```javascript
["America/Dawson", "America/Ensenada", "America/Los_Angeles", "America/Tijuana",
 "America/Vancouver", "America/Whitehorse", "Canada/Pacific", "Canada/Yukon",
 "Mexico/BajaNorte", "PST8PDT", "US/Pacific", "US/Pacific-New"]
```

jsTimeZoneDetect gives the correct result of `America/Los_Angeles`.

[...]

PR #220, opened 2015-05-13
Guess current timezone

timrwood:
This adds the ability to guess the browser's timezone. See #138.

### Demo

http://jsfiddle.net/bvjxt7jr/11/

### Implementation

The implementation is fairly similar to [jstimezonedetect](https://github.com/HenningM/jstimezonedetect). We look through a whitelist of zones and check if the offset matches in both January and June.

There are about 500 different timezone identifiers, and about 60 unique January/June offset pairs in use this year. [The limitations of jstimezonedetect](https://github.com/HenningM/jstimezonedetect#limitations) also apply here. We don't differentiate between `Europe/Berlin` and `Europe/Stockholm` as they are equivalent in recent years.

### Whitelisted names

These are the whitelisted timezone names for each January/June offset pair.

| Jan | Jul | Timezone Name |
| --- | --- | --- |
| -11:00 | -11:00 | Pacific/Pago_Pago |
| -10:00 | -10:00 | Pacific/Honolulu |
| -10:00 | -09:00 | America/Adak |
| -09:30 | -09:30 | Pacific/Marquesas |
| -09:00 | -09:00 | Pacific/Gambier |
| -09:00 | -08:00 | America/Anchorage |
| -08:00 | -08:00 | Pacific/Pitcairn |
| -08:00 | -07:00 | America/Los_Angeles |
| -07:00 | -07:00 | America/Phoenix |
| -07:00 | -06:00 | America/Denver |
| -06:00 | -06:00 | America/Guatemala |
| -06:00 | -05:00 | America/Chicago |
| -05:00 | -05:00 | Pacific/Easter |
| -05:00 | -04:00 | America/New_York |
| -04:30 | -04:30 | America/Caracas |
| -04:00 | -04:00 | America/Santo_Domingo |
| -04:00 | -03:00 | America/Halifax |
| -03:30 | -02:30 | America/St_Johns |
| -03:00 | -04:00 | America/Campo_Grande |
| -03:00 | -03:00 | America/Santiago |
| -03:00 | -02:00 | America/Godthab |
| -02:00 | -03:00 | America/Montevideo |
| -02:00 | -02:00 | America/Noronha |
| -01:00 | -01:00 | Atlantic/Cape_Verde |
| -01:00 | +00:00 | Atlantic/Azores |
| +00:00 | +00:00 | Etc/UTC |
| +00:00 | +01:00 | Europe/London |
| +00:00 | +02:00 | Antarctica/Troll |
| +01:00 | +01:00 | Africa/Lagos |
| +01:00 | +02:00 | Europe/Berlin |
| +02:00 | +01:00 | Africa/Windhoek |
| +02:00 | +02:00 | Africa/Johannesburg |
| +02:00 | +03:00 | Asia/Beirut |
| +03:00 | +03:00 | Europe/Moscow |
| +03:30 | +04:30 | Asia/Tehran |
| +04:00 | +04:00 | Asia/Dubai |
| +04:00 | +05:00 | Asia/Baku |
| +04:30 | +04:30 | Asia/Kabul |
| +05:00 | +05:00 | Asia/Yekaterinburg |
| +05:30 | +05:30 | Asia/Kolkata |
| +05:45 | +05:45 | Asia/Kathmandu |
| +06:00 | +06:00 | Asia/Omsk |
| +06:30 | +06:30 | Asia/Rangoon |
| +07:00 | +07:00 | Asia/Krasnoyarsk |
| +07:00 | +08:00 | Asia/Hovd |
| +08:00 | +08:00 | Asia/Shanghai |
| +08:00 | +09:00 | Asia/Ulaanbaatar |
| +08:45 | +08:45 | Australia/Eucla |
| +09:00 | +09:00 | Asia/Yakutsk |
| +09:30 | +09:30 | Australia/Darwin |
| +10:00 | +10:00 | Australia/Brisbane |
| +10:30 | +09:30 | Australia/Adelaide |
| +11:00 | +10:00 | Australia/Sydney |
| +11:00 | +10:30 | Australia/Lord_Howe |
| +11:00 | +11:00 | Pacific/Noumea |
| +11:30 | +11:30 | Pacific/Norfolk |
| +12:00 | +12:00 | Pacific/Tarawa |
| +13:00 | +12:00 | Pacific/Auckland |
| +13:00 | +13:00 | Pacific/Tongatapu |
| +13:45 | +12:45 | Pacific/Chatham |
| +14:00 | +13:00 | Pacific/Apia |
| +14:00 | +14:00 | Pacific/Kiritimati |

### Keeping whitelist up to date

It is likely that this list will change based on updates to the tzdb, so we will probably need a way to tie this to data updates.

One problem we will need to solve is how to determine which timezone to use in the whitelist for a given offset pair. For example, there are many timezones that match `-08:00/-07:00` this year. `America/Dawson`, `America/Ensenada`, `America/Los_Angeles`, `America/Santa_Isabel`, `America/Tijuana`, `America/Vancouver`, `America/Whitehorse`, `Canada/Pacific`, `Canada/Yukon`, `Mexico/BajaNorte`, `PST8PDT`, `US/Pacific`, `US/Pacific-New`.

The current whitelist is mostly based on [the list from jstimezonedetect](https://github.com/HenningM/jstimezonedetect/blob/a3e06b1f4bbe8d601722d2d274fc93ff3b735421/jstz.js#L279-L350) with a couple additions for new tzdb data. Ideally, we could get data for the population in each timezone and use that to choose the best candidate, but I'm not sure where we could find that data.

### User added whitelisting

Because we are just looking through an array of names, we can expose that list to the user to add their own names.

```js
moment.tz.currentZone(); // America/Chicago
moment.tz.currentZoneWhitelist.unshift('US/Central');
moment.tz.currentZone(); // US/Central
```

We should probably make a dedicated api for this rather than having to know the internals of whether to push or unshift onto the array.

### Bikeshedding

Now the fun part.

There are a few different names we could go with here.

```js
moment.tz.identify();
moment.tz.detect();
moment.tz.guess();
moment.tz.currentZone();
moment.tz.guessZone();
moment.tz.userZone();
```

To add user defined whitelisted names, we also have a couple options.

```js
moment.tz.detect('US/Pacific'); // used as a getter/setter
moment.tz.identify('US/Pacific');
moment.tz.canDetect('US/Pacific');
moment.tz.canGuess('US/Pacific');
moment.tz.currentZoneWhitelist('US/Pacific');
moment.tz.userZoneWhitelist('US/Pacific');
```


timrwood:
For the api names, I'm leaning towards this.

```js
moment.tz.detect(); // get the users timezone
moment.tz.detect('US/Pacific'); // prefer US/Pacific over America/Los_Angeles
moment.tz.detect(['US/Central', 'US/Eastern']); // add multiple preferred zones
moment.tz.ignore('US/Central'); // remove US/Central from the preferred list
```

mj1856:
Still looking over the implementation, but I like the direction this is heading.

Thoughts:
- I'm ok with the `detect` naming, but it sounds very affirmative.  IMHO calling it `guess` would be better aligned with what it does..   Either way, we should call out in the docs that it is an educated guess, not an absolute positive match.  A primary use case would be to pre-select a time zone from a list of time zones - still allowing the user to deviate by choosing a different item in the list.

- Should we prefer the ECMAScript Internationalization API when available?  I made [a similar PR to jsTimeZoneDetect awhile back](https://bitbucket.org/pellepim/jstimezonedetect/pull-request/7/adding-support-for-retrieving-time-zones/diff).

- The Jan/Jul test is a good start, but that doesn't help distinguish between zones with differing DST rules.  The `+02:00` offset zones are particularly problematic.  For example, we have whitelisted `Asia/Beirut` for zones with +2/+3, but that zone's DST transition dates are not aligned with `Asia/Jerusalem`, `Africa/Cairo`, or EU countries like `Europe/Athens` and `Europe/Helsinki`.   jsTimeZoneDetect *attempts* to differentiate between these zones, but [ES5 15.9.1.8](https://bitbucket.org/pellepim/jstimezonedetect/issue/68/effects-of-ecmascript-15918) (our moment/moment#831) makes it difficult.  To do it right is difficult, as it ends up changing from year to year, and we don't have any guarantees about what tzdb updates are on the end-user's computer.  Anyway - do you have any thoughts on this?  Or do we just avoid it and say we limit ourselves to whitelisted zones?

timrwood:
> Should we prefer the ECMAScript Internationalization API when available?

Looks like it is pretty well supported across all but safari, mobile, and ie10 and lower. http://caniuse.com/#search=intl

I'll add a try/catch somewhere that checks if `Intl.DateTimeFormat().resolvedOptions().timeZone` is one of the available zones and returns early.

> To do it right is difficult, as it ends up changing from year to year, and we don't have any guarantees about what tzdb updates are on the end-user's computer.

Yeah, I think this is a pretty slippery slope. I think between framing it as a guess in the documentation and adding the more accurate `Intl` lookup we will have covered a great majority of use cases.

If we did want to try to disambiguate cases like `+02:00`, we could maybe do a binary search on the current year to find the start and end of DST for all matched timezones and filter them from there. It does become difficult to test and maintain those edge cases, especially with data updates, and I'm not sure the cost outweighs the benefits.

blake-nouribekian:
Is this functionality built into the current moment timezone (0.4.0-2015d) yet?

Calyhre:
:shipit:

tests/moment-timezone/guess.js
```diff
@@ -0,0 +1,115 @@
...
+	};
+}
+
+exports.guess = {
+	tearDown : function (done) {
+		Date.prototype.getTimezoneOffset = getTimezoneOffset;
+		done();
+	},
+
+	// These tests were built with 2015b data.
```
sompylasar: Were they built manually or with a script? If a script was written, could you also share it to have the ability to rebuild when `moment` updates with the new data?

moment-timezone.js
```diff
@@ -420,6 +451,32 @@
...
+		for (country in whitelist) {
+			if (whitelist.hasOwnProperty(country)) {
+				zones = whitelist[country].split(' ');
+				for (i = 0; i < zones.length; i++) {
+					guessWhitelist.push(country + '_' + zones[i]);
+				}
+			}
+		}
+	}({
+		africa: 'lagos johannesburg windhoek',
```
sompylasar: Why not e.g. `Africa/Lagos` here (with the upper-case letters and with the slash)?

moment-timezone.js
```diff
@@ -420,6 +451,32 @@
 		momentProperties._z = null;
 	}
 
+	/************************************
+		Guess Whitelist
+	************************************/
+
+	(function (whitelist) {
```
sompylasar: Please add a comment on which version of the tzdb the whitelist was built (the same as in the tests).

sompylasar:
@timrwood Thanks for the great work! Looking forward for this to be merged!

As for the APIs, `browserZone` could be a candidate as this is what actually being guessed.

joshstrange:
Really looking forward to this getting merged! @blake-nouribekian: No, it's not in yet, I wish it was :) I could really use it right about now.

sompylasar:
I haven't got this confirmed for sure yet as it comes in auto-reports from several users of a large online service, and I'm not yet sure if the timezones recorded are set by the web app, but if it is so, then the `Intl.DateTimeFormat().resolvedOptions().timeZone` can return an invalid timezone string, e.g. it returns `"PHT"` instead of `Asia/Manila` in some environments. Would be great if someone could either confirm this or prove it wrong. We've already got "PHT", "COT", "ORAT", "VOLT", "MSK", "TJT", "CST", "IST", "KGT", "IRKT", "YEKT", "GST", "BOT".

We use jsTimezoneDetect as a fallback if the Intl API is missing, and for now we are going to introduce a simple validation around the Intl API to throw away non-existent timezones and fall back to jsTimezoneDetect in these cases, too.

JTallis:
Is this going to be added to momentjs.com soon?

markterrill:
+1

ernestopye:
+1

jeffkole:
+1

Cellule:
+1

mathieumg:
+1 :+1:

phongmedia:
+1 Yes please merge this, it would be great to not have to load an additional third party library to detect user's timezone.

mj1856:
For those +1-ing.  We all understand that client time zone detection is an important use case, but I'm not 100% convinced that this is ready for merging.  There is more analysis to be done.

JTallis:
I've been using this feature. I've had no reports of it reporting an incorrect timezone from various users across the US, UK, Finland, Sweden and Netherlands however that is only a small percentage of all the possible timezones. Just thought I'd say my experience thus far. :)

mj1856:
@JTallis - Thanks, but I wonder...

1. Have you tested near DST boundaries?
2. How prevalent is the actual time zone id in your app?  Would your users notice?
3. Do you have users in Israel, Russia, Africa, South America, etc. that may be getting assigned to EU or US zones and not realizing it?

JTallis:
No, I only have a small user base and it's not really used in anywhere other than one place, which is fairly noticeable. I do also have three other users from Pakistan, Iran and India. I'll ask users to confirm that the timezone chosen for them is correct and if it's not correct, I can report back.

Is there any information you'd need other than the timezone chosen and their actual timezone?

mj1856:
Iran and India are easy to detect due to their fractional offsets.  Pakistan could get detected as a handful of other countries, but I'm not sure it would matter since they don't have DST.

Thanks, but I think the best thing to do is to run the code through testing on different time zones.  We need some grunt scripts that iterate through all time zones, change the system time zone (or env var), run the detection code, and output a list of all matches and mismatches.

Probably best to start with a Linux or Mac environment with the same tzdata update as the one being used in moment-timezone.   Eventually we'd want to expand it to support Windows zones through CLDR mappings.

TWiStErRob:
> We all understand that client time zone detection is an important use case, but I'm not 100% convinced that this is ready for merging. --- https://github.com/moment/moment-timezone/pull/220#issuecomment-120752676

@mj1856 @timrwood
Since this is a best effort guess implementation anyway, wouldn't it be possible to release it as is? The API highly likely won't change and the "best effort" part is an implementation detail that can change over time by becoming "bester" in future releases.

*In any case I bow to your knowledge and efforts. Thank you!*

JTallis:
North Korea's timezone is 30 minutes behind that of South Korea. I'm not sure if changes like this would be "automatically" detected by system updates or not - clueless on that but just thought I'd mention it here just incase.

mj1856:
@JTallis - That change is in tzdb 2015f.  Once merged in, then yes - it would be detected by the script easily, since no other time zone in the world uses the `+08:30` offset.  Of course, their local machine must also be updated accordingly.

mj1856:
@TWiStErRob - I agree in principal, but I'm concerned with opening the flood gates for detection bugs.  I'll leave it to @timrwood to decide.

Since the method is explicitly named `guess`, then I suppose it is ok as long as we are satisfied with the API design, and document it accordingly.

moment-timezone.js
TWiStErRob: +1

pierre-b:
Guys it's possible to narrow down the list of timezones using the navigator.language
Exemple: 
if the guessed list is ['Europe/Berlin', 'Europe/Rome', 'Europe/Paris']
navigator.language == 'fr' => Europe/Paris
navigator.language == 'it' => Europe/Rome
navigator.language == 'de' => Europe/Berlin
etc...
a mapping of timezone/languages is required though

TWiStErRob:
@pierre-b that's a good one, but be careful, [it can contain country as well](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language), which probably narrows it down even more in Europe (`fr-FR`, `fr-BE`, `fr-CA`, `fr-FR`, `fr-LU`, `fr-CH`), but doesn't really help in the USA with `America/*` timezones, because they're all `en-US` or even `es-US`.

I guess using GPS given location to find the nearest point in the map given on http://momentjs.com/timezone/ would be the most accurate.

pierre-b:
@TWiStErRob you are right, USA, Russia and Australia seem to be the only one exceptions. However this mapping reduces the error rate dramatically !

mj1856:
It's usually not a good idea to mix language or locale with time zone.  The primary reason being that people often speak a different language than the inhabitants of where they are physically located.  For example, travelers, immigrants, etc.

I might be a US English speaker living in South America or Europe.  My locale and time zone are two separate things.

mj1856:
That said, since this API is specifically about *guessing* the time zone - when in the case of ambiguous results, we could use the language as another factor for making a more educated guess.  For example, if we've already determined that the user is in a zone that uses UTC+2 in the winter and UTC+3 in the summer, and the locale is `he` or `he-IL`, then the time zone is *probably* `Asia/Jerusalem`.

TWiStErRob:
@mj1856 funnily your previous example would fail: living in Mexico with UTC-7, having `en-US` locale, would give you `America/*` or `US/*` instead of the correct [`Mexico/BajaNorte`](http://www.travelmath.com/time-zone/Mexico/BajaNorte). I guess integrating this into the guess is still mostly helpful, because the number of people with `en-US` living in the USA most likely hugely outnumbers those living elsewhere. (*There's a tradeoff here: big countries have more land, small countries will have correctly guessed timezones* ;))

mj1856:
Yep, the foreigners would always get the short end of the stick in this type of guess.

By the way, you all might be interested to know that timezone detection feature of the [ECMAScript Internationalization API](http://www.ecma-international.org/publications/standards/Ecma-402.htm) has gotten even better with the second edition of the spec.  Eventually, browsers will have to implement this in order to be compliant.

The public API is `Intl.DateTimeFormat().resolvedOptions().timeZone`, but it is optional in the first edition.  Implementers can simply return `undefined.  In the second edition, there is a specific requirement to implement it in section 6.4.3.   :smile: :clap: :joy:  :star2:

timrwood:
I added some more commits that moved the data generation into the build process. I think this was the only thing blocking the initial implementation, we can handle optimizations for locale or Intl based lookups in another pr.

Instead of adding a list of guesses in code, I added a flag at the end of the zone strings.

```js
"Zone/Name_A|PST PDT|80 70|01|-261q0"   // will not be used as a guess
"Zone/Name_B|PST PDT|80 70|01|-261q0|1" // will be used as a guess
```

The specific list of names to be used as guesses is at [tasks/guesses.json](https://github.com/timrwood/moment-timezone/blob/current-timezone/tasks/guesses.json), and is based on [jsTimezoneDetect](https://bitbucket.org/pellepim/jstimezonedetect/src/3a00f59861bddc0d8d169c54dc7b1e7a983b6a3d/jstz.main.js?at=default&fileviewer=file-view-default#jstz.main.js-428).

Unless there are any other concerns, I think this is probably ready to merge.

mj1856:
I updated the demo http://jsfiddle.net/bvjxt7jr/11/

I'll test in a few different time zones, but in general I'm good with this initial step.

timrwood:
@mj1856, once https://github.com/moment/moment-timezone/pull/253 is merged in, I'll rebase on top of develop. Maybe we should release `2015f` as `0.4.1` and this as `0.5.0`?

mj1856:
Yes.  Per semver guidelines, that versioning scheme makes sense.  Plus, I don't want this to hold up a tzdata update.

mj1856:
So, I tested (manually for now) through every time zone on my fully-updated Win 10 machine, under Chrome.  I know the results will be different for different browsers and OSs, but at least I'd like to get an idea of one pass through on a modern platform.

Of the 106 current Windows time zones (comparing against the CLDR mappings), 44 were exactly matched, 6 were matched to acceptable values but not the "001" (golden zone), and 56 were incorrectly detected.   I consider this a 47% success rate.

I think we can do better.  I'm going to go through the failures one by one to see how many can be resolved.  I'd like to get closer to 80% before releasing.

Also, I'd like to be a bit more scientific about how we disambiguate when there's no other option.  We should take into consideration both population estimate, and when possible, [estimated percentage of the population who would have access to Internet technology](https://en.wikipedia.org/wiki/List_of_countries_by_number_of_Internet_users).

For example, there are three Windows zones that are currently in UTC+09:00 without DST.  They map to `Asia/Tokyo` (Japan), `Asia/Seoul` (S. Korea), and `Asia/Yakutsk` (Sakha Republic, Russia).  We currently guess `Asia/Yakutsk` for all three.  We should prefer `Asia/Tokyo`.

We might even be able to isolate `Asia/Yakutsk` in browsers that are already implementing the ES6 or ECMA-402 changes to the specs  (recall moment/moment#831).

**Suggestion:** Can we put this in a new branch?  Then we can merge this and collaborate on improvements before merging back in for 0.5.0.  Thanks.

sompylasar:
> They map to Asia/Tokyo (Japan), Asia/Seoul (S. Korea), and Asia/Yakutsk (Sakha Republic, Russia). We currently guess Asia/Yakutsk for all three. We should prefer Asia/Tokyo.

Please could you add more heuristics like language/locale to avoid such "preference"? And if not, then export all the ambiguous values for the caller to resolve. Working for an international service provider, I can say there is a comparable probability of having a user from Tokyo and from Yakutsk, so such "preference" would leave one of the "ambiguous" users unhappy.

mj1856:
@sompylasar - do keep in mind that the api is called `guess`, not `detect`.  The primary intended use is to pre-select a value in a time zone selection control.  While there may be other use cases, users shouldn't have any delusions about it being perfect.  It *can't* be, until native `Intl` support is perfect in all browsers.

That said, I'd be ok with an API that returns an array of time zones in some approximate order of likelyhood.  Or at least with the primary guess being the first value, so those that don't need the extra data can always just pick the first element.   Or, this could be a different function, or a feature enabled by passing a flag to the guess function.

sompylasar:
@mj1856 - yes, I do!

> That said, I'd be ok with an API that returns an array of time zones in some approximate order of likelyhood. Or at least with the primary guess being the first value, so those that don't need the extra data can always just pick the first element. Or, this could be a different function, or a feature enabled by passing a flag to the guess function.

That's exactly what I meant! The heuristics would help to make the order more precise. Of course, they are not mandatory, but would be very helpful. I understood the original suggestion to "prefer Asia/Tokyo" as "return Asia/Tokyo for this ambiguity", not as "return a list where Asia/Tokyo would be the first item".

timrwood:
> There are about 500 different timezone identifiers, and about 60 unique January/June offset pairs in use this year.

These are the 63 names in https://github.com/timrwood/moment-timezone/blob/current-timezone/tasks/guesses.json

> Of the 106 current Windows time zones (comparing against the CLDR mappings), 44 were exactly matched, 6 were matched to acceptable values but not the "001" (golden zone), and 56 were incorrectly detected.

With the current implementation, we should expect no more than 63 to match. It is curious that there are 19 mismatches. Would you be able to provide the actual name, expected name, and Jun/Jan offsets for those 19?

timrwood:
> I'd be ok with an API that returns an array of time zones

We added lazy loading in https://github.com/moment/moment-timezone/pull/216 to prevent parsing each zone if not necessary. We would probably need to find a solution that doesn't iterate through each of the 500 timezones, effectively undoing the lazy loading optimizations.

Perhaps we could expose an api for adding names to check, and another api to get all guesses that match?

```js
moment.tz.guess()   // 'Asia/Tokyo'
moment.tz.guesses() // ['Asia/Tokyo']
moment.tz.addGuess('Asia/Seoul')
moment.tz.guess()   // 'Asia/Tokyo'
moment.tz.guesses() // ['Asia/Tokyo', 'Asia/Seoul']
```

> Also, I'd like to be a bit more scientific about how we disambiguate when there's no other option.

The list [here](https://github.com/timrwood/moment-timezone/blob/current-timezone/tasks/guesses.json) is a first pass, and we can definitely find ways to optimize it.

mj1856:
Here are my results in a big-ass markdown table: https://gist.github.com/mj1856/7d55b63631ea067bd8b5

Note that by *mismatch* I mean that I don't get the desired mapping.  I recognize that all of the matches are being done by testing two offsets only, so this makes sense.  There's no bug in the current code, it's just not currently as smart as I'd like it to be.

We can adjust the ordering, but I think we can also improve on the algorithm.  We have all of the TZDB data, so rather than just testing two offsets, we could test all of the transitions in the current year and see if they differ.

Additionally, for most of the zones where we don't detect any transitions at all, we may just want to prefer the `Etc/GMT*` zones as not to imply any particular location.

mj1856:
@timrwood - FYI, I'm putting some considerable work into this at the moment.  I'll update this thread shortly.

(Forget about the gist I posted, I'm being more diligent and covering many other cases also).

mj1856:
Ok, I'm done with the first attempt.  For now, it's at: https://gist.github.com/mj1856/e39c77151008e70348d3

This works on a much more comprehensive approach, testing both offsets and transition points.  It also uses ECMA-402 `Intl` if available and working.  It also takes into account the ES5/15.9.1.8 spec bug.

Just load this script in addition to moment-timezone (and data), then call `guessTimeZone()`.  You can also call `guessTimeZone(true)` if you want to bypass `Intl` to see what would otherwise be detected.

The compromise I've made is that when there are no transitions detected whatsoever, either because they don't exist or because the environment hides them, then it returns fixed-offset zones such as `Etc/GMT+7` - *unless* the offset is used *only* in a single location.

For example, for the three I mentioned earlier, `Asia/Yakutsk` will always be detected exactly, but `Asia/Tokyo` and `Asia/Seoul` will only come back exactly if `Intl` is supported, and otherwise will come back as `Etc/GMT-9` instead of picking one haphazardly.

@timrwood - If you're good with my approach, I'll merge more tightly with moment-timezone and open a new PR to replace this one.

mj1856:
And a working jsFiddle: http://jsfiddle.net/rs6hdh5z/

mathieumg:
I'm in Montreal and I get:

```
America/New_York (Intl)
America/Port-au-Prince (Without Intl)
```

Shouldn't I be getting `Etc/GMT-4` for the second one? (EDT)

mj1856:
@mathieumg OS, system time zone setting, and browser please?  Also, in the dev console, you should see a debug output of Candidate and their scores.  Please copy that as well.

No, I wouldn't ever expect an `Etc/GMT*` response from any zone that uses daylight saving time.  I'd actually expect `America/New_York` here, which is good that `Intl` picks it up, but I'm wondering why it thinks it would otherwise be `America/Port-au-Prince`.

mathieumg:
Windows 7 and Chrome 45

image

```
Candidates: America/Port-au-Prince(7.0000000000000036),America/Kentucky/Louisville(6.600000000000005),America/Toronto(6.600000000000005),America/Kentucky/Monticello(6.600000000000005),America/Detroit(6.600000000000005),America/Louisville(6.600000000000005),America/Indiana/Indianapolis(6.600000000000005),America/Montreal(6.600000000000005),America/Indiana/Petersburg(6.600000000000005),America/Nassau(6.600000000000005),America/Indiana/Vincennes(6.600000000000005),America/New_York(6.600000000000005),America/Indianapolis(6.600000000000005),America/Fort_Wayne(6.600000000000005),America/Thunder_Bay(6.600000000000005),America/Pangnirtung(6.600000000000005),America/Nipigon(6.600000000000005),America/Indiana/Marengo(6.600000000000005),America/Iqaluit(6.600000000000005),America/Indiana/Winamac(6.600000000000005),America/Indiana/Vevay(6.600000000000005),America/Grand_Turk(3.2999999999999994)
```

Does it use language information to help a guess?

My languages: `en-US,en;q=0.8,fr;q=0.6`

I think French could be mixing things up along with the fact that Montreal is in the same meridian as Port-au-Prince.

One thing perhaps worth noting is that I had to run that fiddle in Chrome's Incognito mode since I use the HTTPS everywhere extension and it couldn't load Moment from the CDN using HTTPS.

TWiStErRob:
@mj1856 I have some similar results: Windows 7 with en-GB Locale, Chrome 45.0.2454.101 m (64-bit), `Accept-Language: en-US,en;q=0.8,hu;q=0.6`
image

```
Europe/Budapest (Intl)
Europe/Paris (Without Intl)
```

There's no candiate listing in console as it returns (I debugged) one of `unambiguousZones`. What do you mean by "unambiguous" because GMT+1 (2 in summer) has many Europe/City zone aliases.

mj1856:
Thanks.  No, it currently doesn't use language.  I'd rather avoid that if possible.

I'll keep investigating, even though the response is currently valid, it may be a hint of a weakness.

In the meantime, I thought I'd share just a few more examples.  All of the below tested on Windows 10, IE11, Chrome 45, FF 40.

---
Windows Time Zone:  "(UTC+03:00) Moscow, St. Petersburg, Volograd (RTZ2)"
Tim's: `Europe/Moscow`
Matt's: `Europe/Moscow` (Chrome & IE) `Etc/GMT-3` (FF)

Windows Time Zone:  "(UTC+03:00) Minsk"
Tim's: `Europe/Moscow`
Matt's: `Europe/Minsk` (Chrome) `Etc/GMT-3` (IE & FF)

Notice that Moscow is detected with either `Intl` or when ES5/15.9.1.8 is not in play, but Minsk is only detected by `Intl`.  Still, better to return `Etc/GMT-3` than Moscow when uncertain (IMHO).
---

Windows Time Zone: "(UTC-03:00) Montevideo"
Tim's: `America/Montevideo`
Matt's: `America/Montevideo` (Chrome, IE & FF)

Windows Time Zone: "(UTC-03:00) Brasilia"
Tim's: `America/Montevideo`
Matt's: `America/Sao_Paulo` (Chrome, IE & FF)

Windows Time Zone: "(UTC-03:00) Santiago
Tim's: `America/Santiago`
Matt's: `America/Santiago` (Chrome & IE) `Etc/GMT+3` (FF)

Windows Time Zone: "(UTC-03:00) Salvador"
Tim's: `America/Santiago`
Matt's: `America/Bahia` (Chrome & IE) `Etc/GMT+3` (FF)

Notice Montevideo is correctly detected in all cases, but is incorrectly assigned to Brasilia in Tim's algorithm.  Brasilia is correctly identified as `America/Sao_Paulo` in all browsers, due to it's unique DST transitions.

A similar thing happens with Salvador being incorrectly detected as Santiago by Tim's algorithm instead of the correct `America/Bahia`.  In this case, Chrome picks it up via `Intl`, and IE picks it up because it doesn't have the ES5 bug.  Firefox has neither so just sees it as `Etc/GMT+3` since DST was cancelled in 2012.

---

Windows Time Zone: "(UTC+02:00) Jerusalem"
Tim's: `Asia/Beirut`
Matt's: `Asia/Jerusalem"` (Chrome, IE & FF)

Windows Time Zone: "(UTC+02:00) Kaliningrad (RTZ 1)"
Tim's: `Africa/Johannesburg`
Matt's: `Asia/Kaliningrad"` (Chrome & IE) `Etc/GMT-2` (FF)

Windows Time Zone: "(UTC+02:00) Jerusalem"
Tim's: `Asia/Beirut`
Matt's: `Asia/Jerusalem"` (Chrome, IE & FF)

There are many others in this bracket and they are also improved.  The good thing is that most of them have unique DST transitions, or in the case of Kaliningrad have a base-offset change.  This gives them a better signature to detect.  Again, Firefox doesn't use Intl tz detection and is still affected by ES5/15.9.1.8, so it doesn't see any transitions at all for Kaliningrad and thus places it at `Etc/GMT-2` - which is better than assigning it to `Africa/Johannesburg` which is about 9000 miles away on another continent. :smile:

renatus:
Tested last JsFiddle (http://jsfiddle.net/rs6hdh5z/2/) against some timezones in Russia, got pretty interesting results. Ubuntu 14.04, Chrome 46.0.2490.52 beta (64-bit), Chromium 45.0.2454.85 Ubuntu 14.04 (64-bit), Firefox 41.0.

Long-discussed Asia / Yakutsk:

Chrome:
Europe/Moscow (Intl) - WRONG
Etc/GMT-9 (Without Intl) - technically WRONG, it should rather be Etc/GMT+9, it's in Eastern hemisphere :)

Chromium:
Asia/Yakutsk (Intl) - CORRECT
Etc/GMT-9 (Without Intl) - technically WRONG

Firefox
Asia/Yakutsk(No Intl) - CORRECT

Europe / Moscow:

Chrome:
Europe/Moscow (Intl) - CORRECT
Etc/GMT-3 (Without Intl) - technically WRONG, should be Etc/GMT+3

Chromium:
Europe/Moscow (Intl) - CORRECT
Etc/GMT-3 (Without Intl) - technically WRONG, should be Etc/GMT+3

Firefox:
Europe/Volgograd(No Intl) - WRONG
Candidates: Europe/Volgograd(4),Europe/Moscow(4),Europe/Samara(1.9),Europe/Minsk(0.7000000000000001),Europe/Simferopol(0.10000000000000014)

mj1856:
@TWiStErRob - Those results look good.  There are four Windows time zones with identical rules for the EU countries in CET/CEST.  `Intl` (when available) can distinguish between the four, mapping them to one of `Europe/Berlin`, `Europe/Budapest`, `Europe/Paris`, or `Europe/Warsaw`.  But without `Intl` support, they all look the same - so we have to pick one of those to be representative.  I chose `Europe/Paris` arbitrarily (though the IAU and IERS are based there, so that has some merit).

renatus:
Matt's: Asia/Kaliningrad" (Chrome & IE) Etc/GMT-2 (FF) is technically wrong as well, Kaliningrad is located in Europe, so it should be Europe/Kaliningrad.

mj1856:
@renatus - Thanks for the results.  I haven't yet tested on Linux yet, but will before opening my own PR (sorry for stepping on Tim's PR here, but seemed the best place given the prior discussion).

WRT `Etc/GMT-9` - that is actually correct.  All results should be valid IANA TZDB identifiers, as used throughout moment-timezone and other systems, and IDs of this form always have their sign inverted from the ISO8601 form that we normally expect.  We'll have to be sure to clearly document this aspect.

BTW - This may be a good reason to return an array instead of a single zone.  However, this could be tricky - I'd have to expand the logic to find all applicable zones - which would be easy in the case of `Etc/GMT-9`, but hard in the case of `Etc/GMT-3` because `Europe/Moscow` may or may not be detected separately depending on the ES5 bug...   Then again, what real-world use case would it serve to return an array? I'd think most folks will pick the first element and discard the rest anyway.

On your specific results, there are two actionable items.

1. `Europe/Moscow` being detected by `Intl` in Chrome for `Asia/Yakutsk` is *very* wrong.  I hadn't thought much about inaccurate responses from `Intl`, but if this is what you're getting then I should do something about that.  One idea would be to run it through the gamut of remaining tests to see if it behaves as reported.  I'll work on that.

2. `Europe/Volograd` being detected for Moscow in Firefox isn't technically wrong, it's just not ideal.  Volograd and Moscow have been in sync since 1992, and they're in the same CLDR metazone.  However, `Europe/Moscow` is clearly a prefered output, so I will add it and few others to the preference list.

mj1856:
@renatus - WRT `Asia/Kaliningrad` vs `Europe/Kaliningrad` - That was a cut/paste error on my part.  It is indeed detected as `Europe/Kaliningrad`.  I updated the text above.  Thanks.

dantman:
> Then again, what real-world use case would it serve to return an array? I'd think most folks will pick the first element and discard the rest anyway.

How about a timezone select input that lists detected timezones first and a separator before listing other timezones.

mj1856:
@dantman - Interesting! I'd love to see that.  Actually, thinking through the *usual* types of selectors I see, the fixed-offset zones are never present.  So yeah, that will be a problem.  I'll work on this aspect more.  Thanks.

mj1856:
Regarding EU zones, also note that EET/EEST has a similar issue.  There are two Windows identical Windows zones, which `Intl` can distinguish between, mapped as `Europe/Bucharest` and `Europe/Kiev`.  I picked `Europe/Bucharest`.

Refer also to the [CLDR Mappings](http://unicode.org/repos/cldr/trunk/common/supplemental/windowsZones.xml) to see all of the other zones that fall into the same groups.

To clarify, the EET/EEST Windows zones are `"FLE Standard Time"` and `"GTB Standard Time"`.

The CET/CEST Windows zones are `"Central Europe Standard Time"`, `"Central European Standard Time"`, `"Romance Standard Time"` and `"W. Europe Standard Time"`  (yes, "W." ...` is wrong in the zone id -  Go MSFT!)

renatus:
Actually I'm in Moscow, so I had to adjust timezone in OS settings to test Asia/Yakutsk. And it caused the problem - Chrome just didn't picked up new settings fully, despite the fact I've closed it and opened again (because it's not being actually stopped in that case, obviously). I've rebooted PC, and everything is fine now:
Asia/Yakutsk (Intl)
Etc/GMT-9 (Without Intl)
And yes, it would be great to document timezones "invertation", as those Etc/GMT-X are a bit counter-intuitive and hard to find in Google - unlike traditional ones.

mj1856:
@renatus - Oh, so the Moscow result was a fluke then? Good to know.

One thing I've found - browsers are all over the place when you change the time zone while they're running.  On Windows, a `WM_TIMECHANGE` message is fired when the time zone changes.  IE picks it up just fine.  Newer Chrome picks it up, but older didn't used to (can't recall the version).  FireFox picks up the message, but only partially updates the DST rules so detection is screwy until it is the browser is restarted.

I've never had to reboot the whole system to pick up a change, just restart the browser.  Not sure what the eventing system looks like on Linux - so YMMV.

mj1856:
@renatus - See the wording on [Wikipedia](https://en.wikipedia.org/wiki/Tz_database#Area) "The special area of Etc...."

And also [in the tzdb sources `etcetera` file](https://github.com/eggert/tz/blob/master/etcetera#L29-L35).

But yes, we should document this as well.

mathieumg:
> I've never had to reboot the whole system to pick up a change, just restart the browser.

That was perhaps because he has a long running Chrome process that manages Chrome Apps like Google Drive and such. (Even though there is also a standalone version that exists, in that case)

renatus:
@mj1856 Yes, now it's clear, that Moscow time in Yakutsk had nothing to do with your code. Surprisingly enough, I can't reveal that Chrome bug again - now it picks up new timezone fully, and you even don't have to close JsFiddle page, you only have to reload it. Strange, of course...

TWiStErRob:
> Then again, what real-world use case would it serve to return an array?

@dantman exactly my thoughts.
@mj1856 there're other use cases, highly related though:
* display a list of Timezones in a dropdown and highlight or otherwise emphasise the detected ones. Then select the first automatically, this way the user can see which ones are suggested for them.
* List all the zones detected and let the user choose, kind of like when you have a typo in Google search and it suggests a different search term for you.

mj1856:
@TWiStErRob - Can't wait to see all the great time zone pickers that are build with this feature! :+1:

timrwood:
Outside `Intl`, I think there are 3 main phases of the algorithm.

1. Inspect Environment
2. Broad Search
3. Narrow Search + Score

**1. Inspect Environment**

This step uses `Date.prototype.getTimezoneOffset` to find out information about the user's timezone.

My version uses only the January and July offsets of the current year. @mj1856's version uses the actual timestamps for when the offsets changed for the 4 years around the current date. I think @mj1856's version is preferable, although we could probably use a binary search like http://jsfiddle.net/1oq7Lazr/ for better performance.

**2. Broad Search**

This step filters the 500 timezones down to a smaller subset. This is important so that we do not reverse the benefits of https://github.com/moment/moment-timezone/pull/216.

My version uses the hardcoded guesses whitelist to filter down to 63 zones. @mj1856's version uses hardcoded regex of prefixes `(America|Africa|Asia|Australia|Europe|Atlantic|Pacific|Indian|Antarctica)` to filter down to 471 zones.

Instead, we could use the offsets stored in the packed data to filter down options.

`America/Los_Angeles|PST PDT PWT PPT|80 70 70 70|...` would be added to a `70` and `80` offset list. (Note that `70` and `80` are minutes in base 60.) Then, after gathering client information, we can find the zones that overlap with the user's offsets. If the user's offsets are `-08:00` and `-07:00`, Los Angeles would be included in the broad search, but if the user's offsets are `-07:00` and `-06:00`, they would not be included.

**3. Narrow Search + Score**

This step compares offsets and timestamps against each zone in the broad search and assigns a score to them.

My version only returns the first match. I prefer @mj1856's version, which scores each zone based on how many of the offsets and timestamps are matched.

The `preferredZones` could be handled with the guess list of this PR. If a zone is in our guess list, it gets an additional point added to its score. This also could help when preferred zones are not a 100% match.

I rebased this branch down to 1 commit and pushed it to the [`guess-timezone`](https://github.com/moment/moment-timezone/tree/guess-timezone) branch on this repo so that we can make PR's against it.

mj1856:
@timrwood - Great summarization.  Thanks.  I've been doing a lot of experimenting, on both Windows and Ubuntu, and have some improvements to the scoring algorithm.

Closing this PR for now.  When ready, I'll send a PR to the new branch.

mathieumg:
Thanks for all the work put in! Please mention it here once you do open that PR, so we can get notified by email. 

Thank you! :smile:

mj1856:
@timrwood - WRT the binary search for transition you linked - it assumes that there will be one (and only one) transition found.  If no transition is found, it returns the end date of the search, and I'm not sure what it will do if more than one transition exists within the search range (ex: [Morocco](http://www.timeanddate.com/time/change/morocco/casablanca?year=2015)).

I might be able to improve on my algorithm, but I'd still want it to find 0-n transitions within a range, or as I currently have it, the [prior/next] `n` transitions that are no more than `x` years apart.

timrwood:
@mj1856, we could tighten the broad search range from 6 months to 1 month, and do a binary search within that range. I'm taking a stab at *Inspect Environment* as well, should have a pr shortly.

timrwood:
PR for the `Inspect Environment` changes is at https://github.com/moment/moment-timezone/pull/256

marcalj:
Hey, when will be able to use this feature? :) thanks!

prusswan:
can't wait!

jstoeffler:
+1 For those who can't wait, there's http://pellepim.bitbucket.org/jstz/ 
But I can't wait for a moment version :)

mj1856:
Yes, jstz is good, but since that already exists, we don't want to rush moment's version of this out the door without it being significantly better.  We made some progress a few weeks back, but I've been tied up since then (I assume @timrwood has been also).  Hopefully I can get some time to focus on this soon and we can get it in the next release.  Hang in there! :smile:

sompylasar:
Already using `jstz`, yet looking forward for the `moment`'s version!

frekele:
My timezone is America/Sao_Paulo but i see America/Montevideo in the result detected.
I will try to use jstz, because I need detect zone name now.
I waiting for next release the Moment, with zone name.

+1

fmsurvey:
I hope `jstz` and `moment-timezone` can eventually be merged. Timezone work is incredibly messy so it is best to concentrate community efforts.

prusswan:
+1 I hope jstz and moment-timezone can eventually be merged. Timezone work is incredibly messy so it is best to concentrate community efforts.

marcalj:
I don't think it's important to correctly detect the region or zone in the world/map. The timezone is important for the TIME setting. If I'm in Madrid but the browser detects "Europe/London" instead of "Europe/Madrid" is ok because it's the same hour right now.
If I need more accurate location tracking I would use a GPS service or something, but you can never rely on this guesses for tracking.
Thanks!

TWiStErRob:
It's the same hour, but there may be cases (I didn't read the whole data set) when it's currently the same time, but the rules did or will change. Imagine for example if it detects "Europe/London" for you, that is saved in settings and then next year the UK shift DST by one week; or worse: decides to go off DST entirely, then half a year you would have the wrong times. It's all theoretical and I can't give an example, but you get the point.

marcalj:
Yes, but it's something you cannot save it. It doesn't make any sense. If you need the timezone of the user, just get it and send it by parameter.
It's the same if I register in San Francisco and then come back to Europe or some other timezone. It will be incorrect.
I use to get the timezone with ```jstz``` every time I need it. And then send it to server with a custom header.
PD: Also if I change time in my OS, the saved setting will be incorrect. You get the idea.

Thank!

jstoeffler:
What you said makes sense, but it only affects specific cases (user changing the time, or moving to another country) -> it's the user's responsibility and it will probably be less frequent.

Plus you can detect time zone changes.

That's why they call it "Guess", it's good if it's not really sensitive. And if it is, we just have to ask the users, and make sure they understand that if their situation change they have to update (that's what most websites do)

marcalj:
Since it's something that you configure in your OS (sometimes it's automatic, also on mobile), in my perspective it should be transparent and automatic to users.

danielgindi:
You guys know about https://bitbucket.org/pellepim/ right?

TWiStErRob:
That's jstz, it was literally mentioned 3 comments above you... Try `CTRL+F, jstz`.

danielgindi:
Oh yeah, sorry... I read many comments and then just skipped to the end :)

timrwood:
This has been released in `moment-timezone@0.5.0` as `moment.tz.guess()`.

jstoeffler:
Congratulations!

Shobana16:
Super. Thanks for these functionalities. But can someone suggest what exact i can use to get timezone ? i tried guess and zone  but i didnt get any results. actually i am passing a string "Tue, 05 Jan 2016 18:51:18 PST", and trying to get timezone from this to stop converting to any other timezone by checking it. can u pls tell what to use?

danielgindi:
@Shobana16 you are using it incorrectly. You should not pass anything - just call it, and get the current timezone string.

Shobana16:
I am getting this string as an input from other source. so i am trying to get that timezone.

mathieumg:
@Shobana16 You probably want to make a new Moment instance with that string:

```js
console.log(moment(new Date("Tue, 05 Jan 2016 18:51:18 PST")).guess());
```

Shobana16:
ok But i get guess is not a function. :( i have moment version 0.5. do i need any other lib?

mathieumg:
You need moment (latest version is 2.11) as well as moment-timezone version 0.5.

Shobana16:
yes i have that only.

TWiStErRob:
The last comment from Tim (author) was that it was released in 0.5 and he gave the exact usage: `moment.tz.guess()`. It is a "static" function no need to create a Moment object. It gives you what moment-tz can figure out the user's time zone is, not parsing a date string.

I think what you're looking for is significantly different than the purpose of the feature discussed here, I think you should ask a StackOverflow question to get help to prevent further hijacking this issue. From what I understood you want something ike this:
```javascript
moment(new Date(yourString)).tz("America/Los_Angeles").format("ddd, DD MMM YYYY HH:mm:ss z") == yourString
```

mathieumg:
Thanks @TWiStErRob for clarifying, my moment is a bit rusty! I agree that a SO post would be a better place to seek help with this.

mj1856:
The `moment.tz.guess()` functionality is *only* about guessing the user's time zone.  It is not at all about guessing the time zone from any sort of string input.

PR #234, opened 2015-07-22
Specify jspm dependencies for moment

lookfirst:


blixt:
For me this change increased the minified source of my project by almost 200 kB. It seems that for some reason JSPM is now pulling in a long chain of Node.js shim libraries (base64, zlib, buffer, stream, process, …) and Browserify modules, which clearly aren't necessary for moment to work. Any ideas on how to fix this?

lookfirst:
That makes no sense... this change just makes moment-timezone depend on moment. moment doesn't have any other dependencies...

blixt:
@lookfirst Try the following steps in an empty directory and you'll see a big difference between 0.4.0 and 0.4.1:

```bash
$ npm init  # just use all the defaults
$ npm install jspm --save-dev
$ jspm init  # just use all the defaults
$ echo "import moment from 'moment-timezone'; console.log(moment().format());" > index.js
# First, let's try v0.4.0
$ jspm install npm:moment-timezone@0.4.0
$ jspm bundle-sfx index index.bundled.js
$ ls -la
# Now, let's try v0.4.1
$ jspm install npm:moment-timezone@0.4.1
$ jspm bundle-sfx index index.bundled.js
$ ls -la
```

When following the above steps, I see the bundle size going from 105 kB (0.4.0) to 255 kB (0.4.1).

lookfirst:
@blixt I'd file a bug with jspm... that just seems wrong... =(

blixt:
@lookfirst Ok I'll file an issue. By the way, why does the dependency specify `~2.6.0` when that version is one and a half years old? Shouldn't the dependency be `^2.10.6` instead to account for current and future minor versions?

blixt:
https://github.com/jspm/jspm-cli/issues/1193

blixt:
@lookfirst I've been debugging this and I don't think it's jspm's fault. The main reason for the bump in the package size actually seems to be because of #237 which includes the `builds` directory, which then always gets bundled into the JSPM package no matter if you wanted all the data or not.

lookfirst:
@blixt meh. =( well at least it wasn't my change. =)

PR #237, opened 2015-08-05
Fix #195: Include builds dir in npm package

### LINKED ISSUE #195: Compiled files aren't included via npm.
It's really nice to be able to use npm to install dependencies even if you don't plan to bundle them in with requires. Quite frequently I will use a build tool to just concat my core dependencies and bring that in before my app code.

thatmarvin:
This makes it possible for people who use npm for browser-side package
management to use moment-timezone directly without having to build it.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -9,7 +9,6 @@ contributing.md
 .travis.yml
 .jshintrc
 
-builds
 bower.json
 
 data/unpacked/*.json


eddywashere:
:+1: this would be incredibly useful for me

swissmanu:
I'd really like to have the builds included when using NPM. Any chance to get it merged?

PR #256, opened 2015-10-07
Add user offset to help with guesses.

timrwood:
This addresses the `1. Inspect Environment` part of https://github.com/moment/moment-timezone/pull/220#issuecomment-146251011.

It loops over each month in the 4 years surrounding the current date and builds an array of `OffsetAt` objects for each time the offset changed. Within a month, it uses a binary search to find the exact minute the offset changed.

The number of lookups is fairly consistent across timezones. There are a minimum of 48 lookups (1 per month), and for each month that has a change in offset, only ~16 more lookups are needed to find the exact time of change. This results in ~180 lookups for timezones with DST, and 315 for a more complicated zone like Morocco.

This does not address `2. Broad Search`, or `3. Narrow Search + Score`, but those can be addressed in `addZone` and `rebuildGuess` respectively.

cc @mj1856

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -29,6 +29,7 @@
 		links = {},
 		names = {},
 		guesses = [],
+		cachedGuess,
 
 		momentVersion = moment.version.split('.'),
 		major = +momentVersion[0],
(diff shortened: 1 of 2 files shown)


timrwood:
Merging this in to allow further progress on `2. Broad Search` and `3. Narrow Search + Score`.

mj1856:
Thanks!

PR #272, opened 2015-11-25
Fixes #248 - Change return value to always be 'this' to support chaining

### LINKED ISSUE #248: moment().tz(...) return value is inconsistent with moment function chaining.
All moment() methods return `this`, so the method calls can be chained. moment().tz(zoneName) returns the new zone name if it was set (https://github.com/moment/moment-timezone/blob/develop/moment-timezone.js#L383).

rclanan:


mj1856:
LGTM

timrwood:
The current implementation already mirrors the other moment methods. Calling `moment#tz` with a string changes the zone and returns the moment to support chaining. Calling `moment#tz` with no arguments gets the moment's timezone and returns its name.

This is consistent with other methods.

```js
moment().year(2012) // a moment in 2012
moment().year()     // 2012
moment().locale('fr') // a moment in french
moment().locale()     // 'fr'
moment().tz('America/Chicago') // a moment in chicago
moment().tz()                  // 'America/Chicago'
```

Always returning `this` from the `tz` method would cause us to lose the ability to get the moment's timezone.

mj1856:
Not sure how I missed that, but I agree with @timrwood.

PR #274, opened 2015-11-30
Guess timezone by population

timrwood:
This adds a few improvements on guessing timezones.

All current timezones now have a rough population estimate to help as a tiebreaker. The populations are based on https://en.wikipedia.org/wiki/List_of_urban_areas_by_population and googling `X population`. We are not incredibly concerned with accuracy here, one digit of precision is usually enough to differentiate between two similar zones.
2. The timezone's abbreviation is taken into consideration if possible.
3. The scoring algorithm is more granular. We take a handful of samples of the offsets and abbrs in the 4 years surrounding the current year using the binary search from #256. We then compare each zone's offsets and abbrs at those times with the browser reported offsets and abbrs.
4. Sorting is done first by offset differences in minutes, then by differences in abbrs, then by population. The end result is that we should be able to distinguish ~190 zones (see [this commit](https://github.com/timrwood/moment-timezone/commit/ca04433d8454337e5c2a336fe715b659449d6341)) with browser abbr information, and ~60 without browser abbr information.

The next step after this is merged is to pre-filter guesses [here](https://github.com/timrwood/moment-timezone/blob/a884221564914b62981fad0d93038e96c4bf3ad8/moment-timezone.js#L333) so that we are not unpacking every zone when guessing the zone.

cc @mj1856


mj1856:
Acknowledged, but I'm tied up for a bit.  May be some time before I can get to reviewing.  Thanks.

timrwood:
I'm going to merge this in order to continue progress.

It will still be on the `guess-timezone` branch, so there is still time for review before merging into master.

PR #275, opened 2015-12-02
Fix issue when an element with ID of `exports` is defined in the browser.

d10:
Elements with IDs automatically have their IDs exposed. This causes errors where `module` is not defined, and yet `module.exports = moment-timezone` is run.

If possible, could you release this as a patch release?

### DIFF
--- a/moment-timezone-utils.js
+++ b/moment-timezone-utils.js
@@ -10,7 +10,7 @@
 	/*global define*/
 	if (typeof define === 'function' && define.amd) {
 		define(['moment'], factory);                 // AMD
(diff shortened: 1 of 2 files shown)


d10:
Ping any updates?

timrwood:
This export pattern is based on UMD, could you update your PR to use this check instead?

https://github.com/umdjs/umd/blob/f208d385768ed30cd0025d5415997075345cd1c0/templates/returnExportsGlobal.js#L23

d10:
@timrwood updated.

timrwood:
Thanks!

PR #279, opened 2015-12-10
Filter by known offsets before guessing timezone

timrwood:
This is the last step for guessing timezones.

The list of 600 zones is filtered down to the 20 or so that are applicable to the user's offsets. Then, when we get a list of zones to check against, we only have to check against that small handful.

I'd really like to get this feature out the door, so I'll be merging this in over the weekend, and plan to merge the `guess-timezone` branch into develop and release it next Tuesday. We can always improve the algorithm once it is released.

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -28,7 +28,7 @@
 		zones = {},
 		links = {},
 		names = {},
-		guesses = [],
+		guesses = {},
 		cachedGuess,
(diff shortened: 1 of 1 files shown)


mj1856:
Thanks for working on this!  Sory I've been a bit busy lately.  I agree, merge soon and we can always improve later.

Issue #269, opened 2015-11-07
jspm dependency range differs that results in fork


maximnaidenov:
Hi,

It is really nice that you provide jspm overrides and particularly the main and shim settings. But there is a mismatch between the version range for `moment` dependency. In main dependencies section you have >=2.6.0 but then in 'jspm.dependency' you have: '~2.6.0' that resolves to '>=2.6.0 <2.7.0'.
And the problem happens if someone wans to use moment >2.6 because then jspm could not automatically satisfy moment-timezone constraints and so moment is forked and you get two versions being loaded in runtime.
By the way, up to my understanding you don't really need explicit second definition of dependencies in jspm section. jspm-cli resolves them fine in case of npm package from main dependency section.
Please fix the version range or remove the duplicate dependency declaration.

Maxim

gilmoreorless:
Fixed by #284

PR #280, opened 2015-12-11
Fixes/incorrect minimum moment version

kayhadrin:
Fix the moment library dependency version to match with Bower.

This will fix issue #269.


timrwood:
Closing in favor of removing the jspm specific deps altogether in https://github.com/moment/moment-timezone/pull/284.

Issue #290, opened 2016-01-14
GMT+03:00 Asia/Baghdad instead of Europe/Moscow


alealex:
I have 
moment.js version : 2.11.1 
moment-timezone.js version : 0.5.0
os windows 7

If I set timezone to UTC+3 Moscow, St.Petersburg, Volgograd (RTZ 2) then the following code
`"GMT" + moment().format('Z') + ":" + moment.tz.guess()`
will produce `GMT+03:00:Asia/Baghdad`

expected result is `GMT+03:00:Europe/Moscow`

http://jsfiddle.net/3javhx7k/2/

navigator.appVersion is "5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/40.0.2214.91 Safari/537.36"

mj1856:
@timrwood - While some guesses will certainly be wrong, in this particular case we can fix the problem by using the `Intl` api when available.

sompylasar:
Be careful, the `Intl` API returns `Asia/Riyadh` in Chrome, so we cannot actually rely on the `Intl` API, that being a shame for Chrome, but nevertheless.
```
> Intl.DateTimeFormat().resolvedOptions().timeZone
"Asia/Riyadh"
```

Note that none of the guessed candidates in `zoneScores` is `Europe/Moscow`, so there's something wrong with guessing.

sompylasar:
http://pellepim.bitbucket.org/jstz/ did this right somehow, resolves to `Europe/Moscow`.

mj1856:
Intl APIs in newer Chrome does work now.  They used to resolve incorrectly, but that was fixed awhile back (not sure exactly when).

Also, I tested against latest (0.5.13) in IE - which uses the guessing algorithm - and it does now return `Europe/Moscow`.

sompylasar:
@mj1856 Chrome bug here but marked as Started, not Fixed as of now. https://bugs.chromium.org/p/v8/issues/detail?id=3547

mj1856:
@sompylasar  - Yeah, that bug is still present.  However, I don't think it affects our guessing algorithm since Chrome still correctly returns `Europe/Moscow` for the default time zone via `Intl`.

PR #291, opened 2016-01-15
Use Intl API for guessing when available

mj1856:
Since the ECMA-402 Internationalization API provides a mechanism to get the current time zone, we should use it when available and fully implemented.

Note that we also need to guard against a few bad android implementations that return non-IANA time zone IDs.  (jstzdetect ran into this)

This will address #290

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -344,6 +344,10 @@
 		return cachedGuess;
 	}
 
+	function isFunction(input) {
+		return input instanceof Function || Object.prototype.toString.call(input) === '[object Function]';
+	}
+
 	/************************************
 		Global Methods
 	************************************/
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -318,6 +318,15 @@
 	}
 
 	function rebuildGuess () {
+
+		// use Intl API when available and returning valid time zone
+		if (typeof Intl === 'object' && isFunction(Intl.DateTimeFormat)) {
+			var zone = Intl.DateTimeFormat().resolvedOptions().timeZone;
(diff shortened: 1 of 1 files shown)


maxnikulin:
Perhaps an additional sanity check is required. Moment-timezone may have newer timezone info than the system due to the device vendor does not provide updates. That is why another timezone may be set by the user to get correct time offset.

```javascript
var mLocal = moment();
var mGuess = mLocal.clone().tz(zone);
if (mLocal.hours() === mGuess.hours() && mLocal.minutes() === mGuess.minutes()) {
   return zone;
}
```

Also I would suggest to add try-catch.

maxnikulin:
Sorry, just realized that it is more complicated. After moment.tz.setDefault(...) my suggestion will not work.

moment-timezone.js
```diff
@@ -318,6 +318,15 @@
 	}
 
 	function rebuildGuess () {
+
+		// use Intl API when available and returning valid time zone
+		if (typeof Intl === 'object' && isFunction(Intl.DateTimeFormat)) {
+			var zone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+			if (zone && (zone.indexOf("/") > -1 || zone === 'UTC')) {
```
timrwood: How about something like `if (typeof zone === 'string' && names[normalizeName(zone)]) {` to make sure a identifier returned from `Intl` actually has data loaded? Maybe we could add a `logError` if `Intl` returns an identifier that was not loaded in the data set?

mj1856: Yes, I like that idea.  I'll work on that.

timrwood: I continued this work in #304

Issue #294, opened 2016-01-26
moment.tz.guess() error


fritz-c:
Running `moment.tz.guess()` produces the following error:
```js
Uncaught TypeError: Cannot read property 'join' of null(…)OffsetAt @ global.js:11363
userOffsets @ global.js:11408
rebuildGuess @ global.js:11475
guess @ global.js:11496
(anonymous function) @ VM762:2
InjectedScript._evaluateOn @ VM757:875
InjectedScript._evaluateAndWrap @ VM757:808
InjectedScript.evaluate @ VM757:664
```

__OS__: Windows 7, Build 7601
__Browser__: Chrome 47.0.2526.111 m
__Expected timezone__: Asia/Tokyo

Opening the tests page on the momentjs website with the same environment produces no errors. However, in my code and as well as in the console in the momentjs site, running `moment.tz.guess()` results in the same error.

limpbrains:
One of my users have the same error. I can't reproduce it currently. Last Chrome 48. Firefox wors fine.
```
Uncaught TypeError: Cannot read property 'join' of null
```
moment-timezone.js:209
Vilnius timezone

Pitmov:
Same Error. 
OS: Windows 7, Build 7601
Browser: Chrome 48.0.2564.97 (64-bit) and Opera 25.0.1614.71
OS Timezone: (UTC +3:00) Kaliningrad

Firefox works fine.

Pitmov:
Maybe use try catch in offsetAt function? I suppose that "abbr" variable in function contains invalid date. For example, we can do like that:

```
  function OffsetAt(at) {
      var timeString = at.toTimeString();
      var abbr = timeString.match(/\(.+\)/);
      try {
          if (abbr && abbr[0]) {
              // 17:56:31 GMT-0600 (CST)
              // 17:56:31 GMT-0600 (Central Standard Time)
              abbr = abbr[0].match(/[A-Z]/g).join('');
          } else {
              // 17:56:31 CST
              abbr = timeString.match(/[A-Z]{3,5}/g)[0];
          }
      } catch (e) {
          abbr = undefined;
      }

      if (abbr === 'GMT') {
          abbr = undefined;
      }

      this.at = +at;
      this.abbr = abbr;
      this.offset = at.getTimezoneOffset();
  }
```

BerkeleyTrue:
We have over a dozen users reporting this error. Is there a timeline for a fix on this?

[...]

PR #302, opened 2016-02-19
Ensure timezone abbreviations are in alphabets

beata:
The timezone abbreviation(Taiwan or China) on "Chrome / Windows 7" isn't an alphabet string.

It returns something like this:

```
(new Date()).toTimeString()
=> '17:56:31 GMT+0800 (台北標準時間)'
```

I've changed the regular expression in `OffsetAt` to let `abbr` property from non-alphabet timezones   as `undefined`

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -202,13 +202,14 @@
 
 	function OffsetAt(at) {
 		var timeString = at.toTimeString();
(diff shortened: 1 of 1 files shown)


timrwood:
Nice, thanks for the contribution! Looks like this should help address #294.

Issue #298, opened 2016-02-01
Move to ES6 modules to match moment


mattjohnsonpint:
Moment moved to an ES6 module system some time ago.  Moment-timezone should be updated similarly.

timrwood:
I've started work on this in https://github.com/timrwood/moment-timezone/tree/es6-source if anyone wants to follow along.

gilmoreorless:
Closing this off as it's not going to happen in maintenance mode.

PR #303, opened 2016-02-23
Handle uncommon output from Date#toTimeString.

timrwood:
Addresses #294


dominicbarnes:
Just curious, what is the expected output going to be now? Is it just going to return `null` or `false`? Or is something going to be chosen as the default timezone? (like "UTC")

In either case, can the tests being added here cover what the return value should be?

timrwood:
The expected output is that `OffsetAt#abbr` is `undefined` when we were unable to find an abbreviation from the builtin Date object.

A test would be good, but this class isn't exposed publicly at all, so it currently isn't possible to test the exact output. Moving to ES6 modules as noted in #298 would make explicit tests for this possible.

PR #348, opened 2016-05-25
Check for timestamp formats when parsing

### LINKED ISSUE #347: Unix timestamps are parsed incorrectly when using 'x' or 'X' with moment.tz
When parsing a unix timestamp (milliseconds since the epoch), `moment.tz` seems to parse the timestamp, then modify it by applying an offset? At least I think that's what is happening. Example:

``` js
moment.tz(1464194866106, 'x', true, 'America/New_York').format('x');
```

I would expect this to simply output the original timestamp, but it actually outputs `1464209266106`, which is a 240 minute difference (or, the timezone offset for the given timezone).

On the other hand, this:

``` js
moment(1464194866106, 'x', true).tz('America/New_York').format('x');
```

outputs the correct and expected value of `1464194866106`. It's late, so I feel like I must be missing something here. But a timestamp is a timestamp -- so parsing it in the context of a timezone shouldn't change the underlying timestamp value in any way..

joekrill:
Unix timestamps were being parsed incorrectly when using 'x' or 'X' with `moment.tz`. The offset was being applied to the timestamp after the fact, actually changing the underlying timestamp. Fixes #347.

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -464,7 +464,8 @@
 	}
 
 	function needsOffset (m) {
-		return !!(m._a && (m._tzm === undefined));
+		var isUnixTimestamp = (m._f === 'X' || m._f === 'x');
+		return !!(m._a && (m._tzm === undefined) && !isUnixTimestamp);
 	}
 
 	function logError (message) {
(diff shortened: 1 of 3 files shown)


mj1856:
Sorry for the delay.  I'll review this soon. Thanks for your contribution!

maggiepint:
So, we are on a massively slow release cycle. That said, I think your change is correct. Is there any chance I can convince you to sign a CLA for this?

joekrill:
Of course, no problem. Done!

PR #377, opened 2016-09-01
Issue 289; Adding country functionality

alexgrace95:
Added country functionality.
- "grunt data" adds countries to packed and unpacked data from meta data
- Now can parse either Zone or Country code, if country, will return an array with a moment for each zone within the country. Did consider adding all moments methods to array methods however decided not to because it is bad practice and the user can apply the moment methods to the array using .map . All of the moments within the array will be of different zones within the country however some of the zones may have the same time, it is for the user to filter the ones they want or find the unique times.
- Added user endpoints added such as: tz._countries, tz.addCountry, tz.addZonetoCountry, tz.removeZonefromCountry, tz.country and tz.unpackCountry.
- tz._countries is similar to tz._zones and is where all the country data can be accessed,
- tz.addCountry is used to add a new country(should not be needed as all IANA countries exist),
- tz.addZonetoCountry is to add a new zone to an existing country,
- tz.removeZonefromCountry is to remove a Zone from a country,
- tz.unpackCountry unpacks a packed country string into a country object,
- tz.country (similar to tz.zone) gets the specified country information,
- Country object contains country two letter ISO-3166 code, country full name and the zones in the country.
- Have updated and added new tests to reflect the country functionality. All previous tests still work  and there should be no compatibility issues.
- Only accepts two letter country codes and not country full name.

For country codes see here: https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2

If any more functionality is required, please do not hesitate to email or phone me and I will add it.

This is my first pull request and I appreciate any feedback.

Have fun using the country functionality and I hope it is useful and helpful 👍


mj1856:
Thanks.  I'll review tis as soon as I can.

mj1856:
There are multiple places where the following need to be cleaned up:

- Use `===` and `!==` instead of `==` and `!=` unless you expect and desire to work with truthy/falsy values, type coercions, etc.  In most cases you won't.
- Instead of `if (typeof foo === "undefined")`, you probably can use `if (!foo)` in most cases. (Here, falsiness works to your advantage)
- In several places you build arrays from other arrays, by iterating over the first array and pushing to the second.  These can be much cleaner if rewritten to use the `map` function.

More comments pending...

moment-timezone-utils.js
```diff
@@ -219,7 +247,135 @@
...
+				location = location_array[location_array.length - 1],
+				location_space = location.replace("_"," ");
+
+			if (typeof meta.zones[name] != "undefined") {
+				for (var meta_country in meta.zones[name].countries) {
+					countries[meta.zones[name].countries[meta_country]].zones.push(name);
+				}
+			}
+
+			///WORK FROM HERE
```
mj1856: remove comment

moment-timezone.js
```diff
@@ -126,7 +132,18 @@
...
+			countries  : countries
+		};
+	}
+
+	function unpackCountry (string) {
+		var data = string.split('|');
+
+		return {
+			name     : data[1],
+			abbr     : data[0],
```
mj1856: Sequence these consistently throughout please.

mj1856:
Thanks for your contribution, but this is a little too much, and I'm not quite understanding the reasoning for all of the API additions here.

- Why does `moment.tz('US')` return an array of 34 moment objects?  We really cannot redefine how moment construction works.  We do not need code that retrieves the current moment in all time zones within a country.  If we supported a country code here, we'd also have to support it during conversion, as in `moment().tz('US')` - which clearly doesn't work because you can't mutate a single instance into an array.

- It seems like the other main API is `moment.tz.country('US')`, which I would expect to return an array of time zone identifiers for the United States.  Instead, I get an object back with an array of zones, but they're not in a usable form (`america_chicgo` instead of `America/Chicago`).

- I don't want to use the full names of countries in this library, because they are in English only, and moment is supposed to be language-neutral.  We don't have a lot of other display strings here, so the English names of the countries should not be part of the data.

Really, all we need is something that gets the time zone identifiers for the country, or tells you what country/countries a time zone ID is associated with.  Anything else is extraneous.  Keep a minimal footprint please.

I'd suggest a simplified API which is:
```js
var zones = moment.tz.zonesForCountry('US'); // returns ['America/Chicago', ' America/New_York', etc...]
var country = moment.tz.zone('America/Los_Angeles').country; // Single country "US"
```

Also, for this purpose, it would be better to use the single country code from `zone.tab` rather than the array of countries in `zone1970.tab`.

Thanks, and I look forward to your revisions.

alexgrace95:
Hi Matt, thanks for the feedback, I will make these changes as soon as I have a little spare time! Hopefully within the next two weeks

alexgrace95:
Pull Request has been submitted and is passing all checks:
#410

Appreciate any feedback!

Issue #397, opened 2016-10-12
Flip Zone.offset, rename as Zone.utcOffset


mj1856:
The `offset` function on the `Zone` class returns positive values west of UTC instead of east like normal.  We had something similar with then `moment.zone` function, which was deprecated in favor of `moment.utcOffset`.

We should do the same with the `Zone` object.  The `offset` function should be deprecated, and a `utcOffset` function should be added, which returns the same value with the sign inverted.

mj1856:
This is fairly trivial.  A PR would be welcome.

Note, we are just talking about the function.  Do not change the `offsets` array, or the nature of the zone data.

acrodrig:
Hi. I am trying to replace the deprecated function, but there is something I do not understand. The following code prints the same offset for both `offset` and `utcOffset`. I was expecting the sign to be different (per comments above). Please advise.

```js
const moment = require("moment-timezone");

let zone = null;

zone = moment.tz.zone("America/New_York");
console.log(zone.offset(Date.now()));
console.log(zone.utcOffset(Date.now()));

zone = moment.tz.zone("Asia/Tokyo");
console.log(zone.offset(Date.now()));
console.log(zone.utcOffset(Date.now()));
```

AleksMeshkov:
@maggiepint, I'm sorry. Maybe I don't get something, but what was the point to approve PR if the feature is not ready yet?
![image](https://user-images.githubusercontent.com/1904894/34952712-b01cf178-fa2b-11e7-9652-c2335f842e94.png)

![image](https://user-images.githubusercontent.com/1904894/34952892-48c1509a-fa2c-11e7-9b24-1b5b65e27259.png)


now I get this warnings everywhere but I can't fix them until momentjs will fix invert sign bug :-(

Alonski:
So is this safe to use or not? Should I keep `offset` or use `utcOffset`?

adrian-gierakowski:
once this gets fixed it will necessitate a major version bump since otherwise `utcOffset` will break for people who expect current behaviour from it. Could project maintainers confirm that this in fact will be the case so I can replace `offset` with  `utcOffset` to get rid of the console pollution? thanks!

Alonski:
Hey maintainers... Any updates one this?

adrian-gierakowski:
Hey maintainers... Any updates one this?
@maggiepint, @ellenaua

JoanYin:
Would like to hear any updates?

RobLSDev:
[...]

PR #398, opened 2016-10-13
 zone.offset has been deprecated in favor of zone.utcOffset

ssskip:
#397


sharikovvladislav:
Why this is not merged?

maggiepint:
Merging. Note that public member does not invert sign. Need to do that later.

PR #417, opened 2016-11-10
Get offset change instants  within a period of time

surfdude75:
It is important when compiling hour depend variables thru years.


ichernev:
OK, yet another implementation of next/previous transition, this one using `_index`. Figure out which one is more appropriate: #431

Issue #266, opened 2015-10-30
Functions to get next/last DST transitions


mj1856:
From [StackOverflow](http://stackoverflow.com/q/33420531/634824).

It would be nice to have `getNextDST` and related functions.

The function should go in moment proper first, where it will need to scan the local environment's time zone.

Then it can be hooked into moment-timezone such that we use the `untils` array of a specific time zone.

grsabreu:
It would be awesome

desnyki:
You could easily achieve this by getting the moment.tz.zone().untils nearest to current UTC time

mj1856:
Updated the original description.  This can all go into moment-timezone, and should use the functions we created as part of the `.guess()` function when working with an regular `moment` object in local mode.

petrgazarov:
Is this still needed? The stack overflow thread no longer exists, so I'm not sure about what the requirements/context are. Otherwise would love to take this task if not too complex (this would be my first PR to moment js). Thanks!!

mj1856:
@petrgazarov - Yes, it's still needed.  The example in the S.O. thread was:

```js
moment().tz('Europe/Paris').getNextDST();
```

I'm not sure that's the exact API that we'd want, but something like that.  A proposal would be welcome.

mnunezNearsoft:
@mj1856 I would like to work on this issue, where can I hook the new function?

getriebesand:
can you add these Functions to the Zone Object? So I can do moment.ts.zone('Europe/Berlin').getNextTransition(date).isDST()
and getPreviousTransition(date)

maggiepint:
Yes, you should be able to hook the zone.

petrgazarov:
There is already a PR for this for those interesting in reviewing: https://github.com/moment/moment-timezone/pull/431 😃 

I followed @mj1856 's [example](https://github.com/moment/moment-timezone/issues/266#issuecomment-256435499), but hooking it up to `Zone` object shouldn't be a big deal.

PR #431, opened 2016-12-13
Add getNextDST() and getLastDST() functions

petrgazarov:
Resolves https://github.com/moment/moment-timezone/issues/266.

Adds `getNextDST()` and `getLastDST()` on `tz` with an option to pass in a date or moment object as an argument (to be used as benchmark when calculating the result). If no argument is given, compares to current time of the `guess()`ed zone.

Would be happy to add docs once code is approved. Thanks!


petrgazarov:
I saw this locally too - not sure why the tests are failing. If you run `:core` only, they pass. When the whole suite is run, `untils` array ends up with only `[ Infinity ]` in it, which makes me think something is not properly loaded.

getriebesand:
please hook is to the Zone Object, and name the functions getNextTransition and getPreviousTransition because not all Transitions are DST Transitions.

drmrbrewer:
What's the status of this feature?  (I'm not too familiar with the way changes are progressed in a library such as this.)  I'd love to see these functions implemented, not only the next DST transition relative to now, but the next one relative to a specified UTC time.  Are they in a beta version or anything?  I could knock something up for my own purposes using `untils` but I'd rather not re-invent the wheel if it has already been invented.

petrgazarov:
Hey @drmrbrewer, sorry about the delay. If you would like to take this on, I believe the feature is complete on my branch. Just need to pass tests. Thanks!

wayofthefuture:
+1

ichernev:
If it's just a for loop over some offsets, it could be done as a binary search (maybe not worth it if the changes are less than 100). I'm not sure the `guess()` call is warranted, but getting next/previous transition from a timestamp is a nice feature. It should be well defined in case it matches the timestamp as well.

Lev-Shapiro:
@ichernev I was searching for a project to contribute to, all these open pull requests weren't published, because there are some problems in the code, right? So that means if someone will fix it - it will be merged?

Issue #343, opened 2016-05-23
Module 'moment' has no exported member 'MomentFormatSpecification'


stefan505:
I'm trying to use moment-timezone in the browser. Installed it like this using `typings 1.0.4`:

```
typings install dt~moment-timezone -SG
```

The resolved `typings\globals\moment-timezone\index.ts` file ends up with these lines that is causing TS build failure:

```
    (date: string, format: moment.MomentFormatSpecification, timezone: string): moment.Moment;
    (date: string, format: moment.MomentFormatSpecification, strict: boolean, timezone: string): moment.Moment;
    (date: string, format: moment.MomentFormatSpecification, language: string, timezone: string): moment.Moment;
    (date: string, format: moment.MomentFormatSpecification, language: string, strict: boolean, timezone: string): moment.Moment;
```

Is this a bug or am I doing something wrong?

DaanDeMeyer:
I have the same problem. The issue is probably that the moment-timezone typescript bindings are not up to date with moment typescript bindings.

mj1856:
See #341 for TBD on adding typescript bindings for moment-timezone.  We'd certainly accept a pull request.

maxkarpovets:
What is pull request number? 

I need it for typings.

konrad-garus:
For now a workaround that let me get past this is:

```
typings install --global dt~moment-node dt~moment-timezone
```

berkeleydiego:
angular2newb here. thanks @konrad-garus for providing that workaround. I gave that a try and it seems to install. However, I get the following run-time error so something still ain't right:

```
[...]

Issue #341, opened 2016-05-10
Add typescript definitions


mattjohnsonpint:
Since moment now has `moment.d.ts`, we should probably add `moment-timezone.d.ts` here.

We can start with this one, and update it:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/moment-timezone/moment-timezone.d.ts

jocull:
+1

jtbairdsr:
+1

marcobisio:
+1

Thiyagarajan-A:
+1

Thiyagarajan-A:
May be this is the updated link:
https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/moment-timezone/

gilmoreorless:
This was done in version `0.5.29`.

PR #310, opened 2016-03-08
Convert to ES6

timrwood:
This addresses #298 by converting the source into smaller es6 modules.

This is not ready to merge yet. Here are the outstanding tasks.

Some of this work can probably be done in parallel. If you want to help out, edit the description or leave a comment about which task you are working on, then open a PR to the [es6](https://github.com/moment/moment-timezone/tree/es6) branch.
- [x] Convert source and tests to es6 modules. @timrwood
- [x] Convert [tests/moment-timezone/default.js](https://github.com/moment/moment-timezone/blob/develop/tests/moment-timezone/default.js) to es6.
- [x] Convert [tests/moment-timezone/manipulate.js](https://github.com/moment/moment-timezone/blob/develop/tests/moment-timezone/manipulate.js) to es6.
- [x] Convert [tests/moment-timezone/parse.js](https://github.com/moment/moment-timezone/blob/develop/tests/moment-timezone/parse.js) to es6.
- [x] Convert [tests/moment-timezone/utc.js](https://github.com/moment/moment-timezone/blob/develop/tests/moment-timezone/utc.js) to es6.
- [x] Update grunt tasks to es6. #311
- [x] Update `grunt data-tests` to generate correct tests. #318
- [x] Update `grunt data-pack` to output `src/data/packed.js` and `src/data/packed-2010-2020.js`. #318
- [x] Move `tasks/population.json` to an es6 module in `src/data/population.js`. #311
- [x] Move `data/meta/latest.json` to `src/data/meta.js`. #318
- [x] Discuss removing `data/**/20*.json` files.
- [ ] Build `src` into `build` for different environments. (es6, globals, amd, commonjs, umd).
- [ ] Point `package.json` and `bower.json` to the correct file entry points.
- [ ] Figure out how to run tests in a browser. sindresorhus/ava#24

PR #444, opened 2017-02-04
add typescript definitions

bmagistro:
Reference: Add typescript definitions #341
Reference: Module 'moment' has no exported member 'MomentFormatSpecification' #343 

I couldn't come up with a solution to 343 other than copying lines 9-13 into the typescript file for moment-timezone.

Supports: https://github.com/grafana/clock-panel/pull/19

### DIFF
--- a/moment-timezone.d.ts
+++ b/moment-timezone.d.ts
@@ -1,0 +1,68 @@
+// Type definitions for moment-timezone.js 0.2
+// Project: http://momentjs.com/timezone/
+// Definitions by: Michel Salib <https://github.com/michelsalib>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+/// <reference path="../moment/moment.d.ts" />
+
+declare namespace moment {
+  interface MomentBuiltinFormat {
+    __momentBuiltinFormatBrand: any;
+  }
+
+  type MomentFormatSpecification = string | MomentBuiltinFormat | (string | MomentBuiltinFormat)[];
+
+    interface Moment {
+        tz(): string;
+        tz(timezone: string): Moment;
+        zoneAbbr(): string;
+        zoneName(): string;
+    }
+
+    interface MomentZone {
+        name: string;
+        abbrs: string[];
+        untils: number[];
+        offsets: number[];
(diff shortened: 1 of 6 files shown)


moment-timezone.d.ts
```diff
@@ -0,0 +1,68 @@
...
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+/// <reference path="../moment/moment.d.ts" />
+
+declare namespace moment {
+  interface MomentBuiltinFormat {
+    __momentBuiltinFormatBrand: any;
+  }
+
+  type MomentFormatSpecification = string | MomentBuiltinFormat | (string | MomentBuiltinFormat)[];
```
westy92: Is the inconsistent indentation intentional?  (This and above vs. the interfaces below.)

westy92: Most of the file is indented correctly--just be consistent.

westy92:
You should also add some tests like what moment has.

See https://github.com/moment/moment/tree/d28a819dc668f107566f306f827df07d43e30287/typing-tests and https://github.com/moment/moment/blob/d28a819dc668f107566f306f827df07d43e30287/Gruntfile.js#L173.

bmagistro:
Not sure if this is something that can be added to the build deps to allow CI to continue, but there is a dependency on "grunt-exec" and "ntypescript".

Tests came directly from DefinitelyTyped.

bmagistro:
Changes to ".travis.yml" align with the changes being requested in #310

bmagistro:
Any updates on this getting merged in?

maggiepint:
So, this situation has been nothing but heartbreak since we started it in the main moment library. Does this package currently work? https://www.npmjs.com/package/@types/moment-timezone

If so, strong preference for leaving them there.

bmagistro:
At this point I don't remember if it worked and I couldn't figure it out or if it didn't work.  I am going to say I likely had an issue with the mentioned package likely similar if not the same as #343.

bmagistro:
Closing PR, trying to redo the update and it appears this was resolved, likely in #343

PR #460, opened 2017-03-22
Updated data to IANA TZDB 2017b

ssskip:


nithin123456:
moment.tz.names() doesn't work?

Eight-Ball:
@mj1856 i'v a similar problem i think : moment.tz is undefined
Any idea ?

pscanlon1:
+1

pscanlon1:
moment.tz undefined here too

davidungio:
Getting the same issue as well: `Uncaught TypeError: moment(...).tz is not a function`

ssskip:
@nithin123456 


image

PR #464, opened 2017-03-25
README - Add module import usage

gnhuy91:


mj1856:
The docs [already describe how to use it](http://momentjs.com/timezone/docs/#/use-it/).  The readme is not the docs.

Also, `import` doesn't work for all environments.

Issue #506, opened 2017-06-29
Different behaviour of parsing function(s)


dentuzhik:
According to the [docs](https://momentjs.com/timezone/docs/#/using-timezones/parsing-in-zone/):
>  Using `moment.tz(Number|Date, zone)` is functionally equivalent to `moment(Number|Date).tz(zone)`

For me this is not true, here's the code which causes the issue:
```javascript
const moment = require('moment-timezone');
const sourceTimeZone = 'Europe/Helsinki';
const targetTimeZone = 'Pacific/Fiji';
const timestamp1 = moment.tz('2017-06-27 18:00', sourceTimeZone).unix();
const timestamp2 = moment('2017-06-27 18:00').tz(sourceTimeZone).unix();

const formatTime = function(timestamp, timezone) {
    return moment.unix(timestamp).tz(timezone).format('h:mma');
}

assert.equal(
    formatTime(timestamp1, targetTimeZone), // should be 3:00am
    formatTime(timestamp2, targetTimeZone)
);
```

If machine, which this test is run on is in the same timezone as `sourceTimeZone` test passes successfully, as soon as I change the timezone of my machine, test fails because formatting of timestamp2 produces different result. I discovered it because test failed in CI for me.

I will attach a PR with a failing test.

gilmoreorless:
I agree the documentation is a bit confusing here, but it is correct. It says that the two behaviours are equivalent only when using numbers or `Date` objects as inputs:

> Unix timestamps and `Date` objects refer to specific points in time, thus it doesn't make sense to use the time zone offset when constructing. Using `moment.tz(Number|Date, zone)` is functionally equivalent to `moment(Number|Date).tz(zone)`.

Your example is using strings, which go through the parsing behaviour to read/convert offsets.

PR #507, opened 2017-06-29
Implement failing test for #506

dentuzhik:
Refer to #506 for explanation

### DIFF
--- a/tests/moment-timezone/parse.js
+++ b/tests/moment-timezone/parse.js
@@ -251,6 +251,23 @@ exports.parse = {
 		t.done();
 	},
 
+    "check that moment(date).tz(timezone) and moment.tz(date, timezone) behave the same way" : function (t) {
+        var sourceTimeZone = 'America/Los_Angeles';
+        var targetTimeZone = 'America/New_York';
+        var timestamp1 = moment.tz('2017-06-27 18:00', sourceTimeZone).unix();
+        var timestamp2 = moment('2017-06-27 18:00').tz(sourceTimeZone).unix();
+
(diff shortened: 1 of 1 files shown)


dentuzhik:
In my default timezone test fails, but if I set it to `America/Los_Angeles` it passes.

tests/moment-timezone/parse.js
```diff
@@ -251,6 +251,23 @@ exports.parse = {
...
+        var targetTimeZone = 'America/New_York';
+        var timestamp1 = moment.tz('2017-06-27 18:00', sourceTimeZone).unix();
+        var timestamp2 = moment('2017-06-27 18:00').tz(sourceTimeZone).unix();
+
+        var formatTime = function(timestamp, timezone) {
+            return moment.unix(timestamp).tz(timezone).format('h:mma');
+        };
+
+        t.equal(
+            formatTime(timestamp1, targetTimeZone), // should be 3:00am
```
dentuzhik: This comment is actually irrelevant, I should remove it

tests/moment-timezone/parse.js
```diff
@@ -251,6 +251,23 @@ exports.parse = {
 		t.done();
 	},
 
+    "check that moment(date).tz(timezone) and moment.tz(date, timezone) behave the same way" : function (t) {
+        var sourceTimeZone = 'America/Los_Angeles';
```
ellenaua: Lot lot lot of people think that 
```moment(date).tz(timezone) and moment.tz(date, timezone)``` 
should work in a same way.
We should definitely update the docs

Issue #423, opened 2016-11-24
Guess should ignore Intl detection bad results


jou:
Sometimes, the Internationalization API doesn't return sensible values. I encountered that in Chrome for Android, where it returns "CET" instead of a timezone ID when the selected system timezone is using Central European Time.

Currently, I'm removing `Intl.DateTimeFormat` before calling `moment.tz.guess(true)` again. That's not really a nice solution, so can an option be added to bypass it?

SuperMario77:
I also vote for such an option. I work on a cordova app on Windows 10 tablet and there the Microsoft Edge Browser Engine will be used. I noticed that the call to "Intl.DateTimeFormat().resolvedOptions().timeZone" always returns a cached value.

Steps to reproduce the problem:
1. Start Microsoft Edge on Window 10 machine
2. Enter Javascript Console by pressing F12 and execute this command:
   Intl.DateTimeFormat().resolvedOptions().timeZone
3. The timezoneId should match the currently set timezone of your system
4. Change your system timezone to another timezone
5. Execute again this command:
  Intl.DateTimeFormat().resolvedOptions().timeZone
6. The timezoneId is still the same from the first call. <= Cached result? Bug?

This procedure works in Chrome without problems. Chrome noticed the changed timezone. So Microsoft has a bug or a "cache feature" ;-) in his Intl.DateTimeFormat() implementation. And because moment-timezone uses this Intl Feature, the moment.tz.guess(true) method also returns the wrong timezone in my cordova app (Edge Browser Engine).

If I remove the Intl Feature at the beginning by doing:
Intl.DateTimeFormat = undefined;

then moment-timezone always guesses the correct timezoneId in my win 10 cordova app.

mj1856:
I'm not concerned about the cached part.  Various browsers have gone either way on that over the years.  If you're concerned about that, raise an issue with Edge / ChakraCore.

I do think it would be good to make sure Intl returned a zone with a `/` char in it (or `UTC`) before accepting the results.  Note, [jsTimeZoneDetect had a similar issue](https://bitbucket.org/pellepim/jstimezonedetect/issues/116/dont-use-abbreviations-to-report-timezone).

For now, you could simply remove (or don't add) the zones that use three-letter abbreviations instead of full identifiers.

mj1856:
To clarify - a true/false option isn't very useful IMHO, but ignoring bad results is.

PR #529, opened 2017-09-20
bypassing chrome intl timezone bug fix #517

### LINKED ISSUE #517: Wrong timezone for China/Schanghai
According to [wiki](https://en.wikipedia.org/wiki/List_of_time_zone_abbreviations)
The abbr. **CST** stands both for **Central Standard Time (North America)** and **China Standard Time**

If you set your timezone on your PC to Shanghai, moment determines timezone: 
`America/Chicago`

**System**
Ubuntu, FF or Chrome
`"moment": "2.18.1", "moment-timezone": "0.5.13"`

chrome-console

timezone

bolasblack:


mj1856:
Thanks, but I don't think this much examination is required.  The reported bad results have all been three-character abbreviations, from what I have been able to ascertain.  Closing in favor of #553

bolasblack:
Hi @mj1856 , I think you misunderstood me, I was trying to fix the issue #517 , not #423

In some region (for example `Asia/Shanghai`), when I executing `Intl.DateTimeFormat().resolvedOptions().timeZone` in chrome 61.0.3163.100, I will get `America/Chicago`, not `Asia/Shanghai`

I think check `Intl.DateTimeFormat().resolvedOptions().timeZone.length > 3` can not fix this issue.

mj1856:
Oh, I see.  Let me take another look at this.  Thanks.

gregclarke:
In addition to the above, in some circumstances the resolved timezone is incorrect. On arch linux firefox version 56.0.2 `Intl.DateTimeFormat().resolvedOptions().timeZone` returns `Etc/GMT-10` which is an hour out due to daylight savings (correct timezone is `Australia/Sydney`). The length check doesn't work in this case either. I've temporarily solved by using
`var tmpName = intlName.split('/')[0]; if (intlName && tmpName.length > 3)`

iffy:
@mj1856 Can you please reopen this/reconsider this as a way to fix 517?

morajabi:
@mj1856 I have this problem :( How can we fix it?

Issue #449, opened 2017-02-23
Error while using moment 


sushant23:
I am getting the following error while trying to use this library in my project 

> Uncaught (in promise) TypeError: Cannot read property 'split' of undefined
>     at Object.eval (moment-timezone.js:34)
>     at eval (moment-timezone.js:12)
>     at eval (moment-timezone.js:18)
>     at Object../node_modules/moment-timezone/moment-timezone.js (reactBoilerplateDeps.dll.js:4931)
>     at __webpack_require__ (reactBoilerplateDeps.dll.js:21)
>     at eval (index.js:1)
>     at Object../node_modules/moment-timezone/index.js (reactBoilerplateDeps.dll.js:4924)
>     at __webpack_require__ (reactBoilerplateDeps.dll.js:21)
>     at eval (eval at 42 (36.chunk.js:83), <anonymous>:1:42)
>     at Object.42 (36.chunk.js:83)

SamMatthewsIsACommonName:
Hi I was having the same issue make sure you are using at least moment 2.6.0 as in the docs

sushant23:
I am currently using latest version i.e. 2.17.1 and I am still getting the same error

BipinBhandari:
This issue seems to be fixed by settings `moment=moment.default` in moment-timezone.js line 20 

```
}(this, function (moment) {
	"use strict";
	moment=moment.default;
```

peterjcaulfield:
I'm also experiencing this issue using es6 imports. 

dep versions:
```
"moment": "2.17.1",
"moment-timezone": "^0.5.11",
```
code:

```
import moment from 'moment';
import 'moment-timezone';
moment.tz.setDefault('Europe/London');
```

Produces:
```
Uncaught TypeError: Cannot read property 'split' of undefined
    at Object.eval (webpack:///./~/moment-timezone/moment-timezone.js?:37)
```

Not 💯 on whether or not I am importing `moment-timezone` correctly but I couldn't find any documentation on how to handle this.

lmammino:
Same issue here, I hope this get addressed any time soon 😞

butterflyhug:
[...]

PR #541, opened 2017-10-10
make compatable with moment 2.19 (as a module)

### LINKED ISSUE #540: the published version is incompatible w/ moment 2.19.0's module format
When importing as an ES module, the umd wrapper fails to account for the possible 'default' export name of moment.js (2.19 added the 'module' property to the package.json, and webpack follows it instead of main.)

jsg2021:
Fixes #540

### DIFF
--- a/moment-timezone-utils.js
+++ b/moment-timezone-utils.js
@@ -6,12 +6,16 @@
 
 (function (root, factory) {
 	"use strict";
+	
+	function interop (exports) {
+		return exports && expoerts.default || exports;
+	}
 
 	/*global define*/
 	if (typeof define === 'function' && define.amd) {
 		define(['moment'], factory);                 // AMD
 	} else if (typeof module === 'object' && module.exports) {
-		module.exports = factory(require('./'));     // Node
+		module.exports = factory(interop(require('./')));     // Node
 	} else {
 		factory(root.moment);                        // Browser
 	}
--- a/moment-timezone-utils.js
+++ b/moment-timezone-utils.js
@@ -6,16 +6,12 @@
 
 (function (root, factory) {
 	"use strict";
-	
-	function interop (exports) {
-		return exports && expoerts.default || exports;
-	}
 
 	/*global define*/
 	if (typeof define === 'function' && define.amd) {
 		define(['moment'], factory);                 // AMD
 	} else if (typeof module === 'object' && module.exports) {
-		module.exports = factory(interop(require('./')));     // Node
+		module.exports = factory(require('./'));     // Node
 	} else {
 		factory(root.moment);                        // Browser
 	}
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -7,11 +7,15 @@
 (function (root, factory) {
 	"use strict";
 
+	function interop (exports) {
+		return exports && expoerts.default || exports;
+	}
+
 	/*global define*/
 	if (typeof define === 'function' && define.amd) {
 		define(['moment'], factory);                 // AMD
 	} else if (typeof module === 'object' && module.exports) {
-		module.exports = factory(require('moment')); // Node
+		module.exports = factory(interop(require('moment'))); // Node & WebPack
 	} else {
 		factory(root.moment);                        // Browser
 	}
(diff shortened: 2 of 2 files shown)


Sieabah:
Hope this gets in soon, none of the solutions I've found in the relevant issues actually seem to fix moment-timezone.

brendonco:
@Sieabah You can try to downgrade moment for now.
https://github.com/moment/moment-timezone/issues/449

Sieabah:
@brendonco I've tried downgrading the versions to 2.18.1 and that didn't work for some reason. Is there a specific version of moment-timezone I need to downgrade as well or is latest alright?

My specific code is

    import * as momentTZ from 'moment-timezone';
    ...
    momentTZ().tz....

jmenglis:
@Sieabah 

You should be able to keep moment-timezone as latest.

I added `"moment": "2.18.1",` to my dependencies in my package.json and ran npm install

jsg2021:
remove the “* as “ part of your import.

Sieabah:
They're functionally equivalent.

jsg2021:
@Sieabah no, they are not. That is a namespace import, not a default import... some libraries this works on, but not this one. This library exports a single function and replaces the `module.exports` with that function. Its subtle, but it matters.

The result of `import * as foo` results in an `Object` with all the named exports as properties. ie:
```js
exports = require(module);
foo = {}; 
for (let key in exports) {
    foo[key] = exports[key];
}
```
This is an overly simple representation, but it should make my point. 

In general for commonjs modules, if they want you to use it like so:
```js
var momentTZ = require('momentTZ')
``` 
you should use this form when translating to ES6/2015+: 
```js
import momentTZ from 'moment-timezone'
```

brendonco:
@Sieabah Have you delete your node_modules folder, npm cache clean --force ? And npm install, once installed. Check your node_modules/moment-timezone/moment or if you're doing what @jmenglis suggested you'll be checking node_modules/moment folder instead. And check the version changelog.md if its 2.18.1. And Node version should be 7++, npm version 5++ in order your npm install to work ;)

Sieabah:
I've done everything recommended and still have this issue. I assume moment-timezone is pulling in moment itself rather than using a peer dependency.

jsg2021:
until this is merged, it will fail for moment@2.19.0+ in es module environments, you will need to specify `"moment":"~2.18.1"` in your app's package.json.

Sieabah:
@jsg2021 Then there is an issue with @types/moment-timezone because doing

`import moment from 'moment-timezone'

Results in an error, which until docker bluescreened me, is something around `moment-timezone does not have a default export`, although it compiles anyway but with that error.

jsg2021:
That is a problem with how whomever authored `@types/moment-timezone`... not this.

Sieabah:
I suspected as much... Ah, whatever.

stantoncbradley:
this branch works for us on react-native 0.49 and moment 2.19.  would love to see this merged so we can upgrade! 😁

mj1856:
Please re-validate with 2.19.1.  Since 2.19.0 had a bug, and that was fixed, I'm not sure that this change is needed now.

jsg2021:
@mj1856 moment offers an es module export. (`module` and `esnext:main` point to an es module) Which means that this library still needs this interop. (that is until this project is rewritten as an es module and compiled to commonjs for those who do not wish to use esm)

stantoncbradley:
`2.19.1` seems compatible with `0.5.13`

"react": "16.0.0-beta.5",
"react-native": "0.49.3",

jsg2021:
i don’t think react-native’s bundler honors `esnext:main` so, in that environment you get the commonjs build... which skirts the issue.

mj1856:
So, unless I'm misunderstanding, this isn't absolutely required.  Given that we do want to repackage/rewrite moment-timezone into ES6 modules before a 1.0 release, I'm going to skip this for now.

Do let me know if there's some scenario I'm not considering.  Thanks.

jsg2021:
Just clarifying... it does work now... however:

rollup and webpack (with [mainFields](https://webpack.js.org/configuration/resolve/#resolve-mainfields) set to look at `esnext:main` along with `module`, and `main`) will still hit this issue. If moment.js adds the `module` field back we will hit this problem.

PR #476, opened 2017-04-06
Resolve es6 loading issue where moment is undefined

brinley:
Fix for https://github.com/moment/moment-timezone/issues/449

PR #551, opened 2017-10-23
Add es module default export interop

### LINKED ISSUE #540: the published version is incompatible w/ moment 2.19.0's module format

aortbals:
Closes #540 and #449. This replaces #541 and #514.

Moment, when imported as an ES6 module, has a default export. This isn't handled by `moment-timezone`. The addition in this PR is the same output you would receive from a UMD build from [rollup](https://rollupjs.org/) that will handle the case that is causing the reported default import issues.

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -18,6 +18,8 @@
 }(this, function (moment) {
 	"use strict";
 
+	moment = moment && moment.hasOwnProperty('default') ? moment['default'] : moment;
+
 	// Do not load moment-timezone a second time.
 	// if (moment.tz !== undefined) {
 	// 	logError('Moment Timezone ' + moment.tz.version + ' was already loaded ' + (moment.tz.dataVersion ? 'with data from ' : 'without any data') + moment.tz.dataVersion);


mj1856:
What scenario specifically does this resolve?  The linked moment issue is about 2.19.0, which had a bug.  It shouldn't be required for 2.19.1.  Not sure on the linked moment-timezone issue.

We do want to repackage/rewrite moment-timezone as an ES6 module eventually.  Is this still required in the meantime?

aortbals:
This resolves a scenario where `jsnext:main`  or `esnext:main` is used to resolve `moment`. This comment also mentions it: https://github.com/moment/moment-timezone/pull/541#issuecomment-339427551.

Here is a [reproduction deployment](https://moment-timezone-551-bxusdiyxpq.now.sh/) showing the reported exception being thrown. Here is the commit that reproduces the error: https://github.com/aortbals/mmt-timezone-551/commit/3233f21f882f33288814b7a6102f865763a51642.

Here is the [fixed deployment](https://moment-timezone-551-riivzddmrj.now.sh/) the version from this PR https://github.com/aortbals/mmt-timezone-551/commit/a350dca4ea9d5c6abd753d00550bd5df1e256806.

petrjaros:
Thanks for making this pull request. I am unable to use https://www.npmjs.com/package/date-holidays without this.

petrjaros:
This pull request is identical with this one: #476

petrhanak:
@mj1856 ES6 modules are nice to have but even without them it's possible to use this library. What we are talking here is bug - using `moment` with `jsnext:main` breaks `moment-timezone`. It's common use case people are struggling with.

Unlike #541 this PR also fixes AMD import (used in certain webpack configuration) and doesn't rely on non-environment properties like #476 does.

ellenaua:
Merged this fix https://github.com/moment/moment-timezone/pull/848/commits/0296382042a7b656f278c828eeaa38872a131e8c

PR #552, opened 2017-10-24
Updated data to IANA TZDB 2017c

ssskip:
1. Updated data to IANA TZDB 2017c
2. Remove Canada/East-Saskatchewan from test case
    

> Changes to zone names
>     Remove Canada/East-Saskatchewan from the 'backward' file, as it
>     exceeded the 14-character limit and was an unused misnomer anyway.


mj1856:
Thanks!

PR #577, opened 2018-02-11
Updated data to IANA TZDB 2018c

ssskip:
testGuess break on:

- America/Cancun - guess
- America/Fort_Nelson - guess
- America/Metlakatla - guess


JohnColvin:
any word on when this will be merged? thanks!

ssskip:
@JohnColvin   since 2018d (Released 2018-03-22),  i need update this pull request then

ellenaua:
@ssskip do you have an idea how to fix failing tests? I see that when `ts.guess()` gets an equal score for multiple time zones, it's choosing zone which is more populated:
```
'America/Phoenix' scores = 0 0 population= 4200000
'America/Fort_Nelson' scores = 0 0 population= 3900
```
That's why instead of 'America/Fort_Nelson'  tz.guess() returns 'America/Phoenix'. And the same happens for 'America/Metlakatla' ('America/Anchorage' is more populated) and  'America/Cancun' ('America/Panama' is more populated). Should we just change test expectations?

ellenaua:
I've done 2018d PR https://github.com/moment/moment-timezone/pull/596

PR #587, opened 2018-03-13
Exclude vulnerable versions of moment

cy6erskunk:
There is a vulnerability in `moment` prior to the 2.19.3
http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-18214


cy6erskunk:
I've just rebased the branch, this tiny update seems to be straightforward and safe

PR #596, opened 2018-04-13
Updated data to IANA TZDB 2018d

ellenaua:
- Updated data to IANA TZDB 2018d
- Fixed 3 failing tests by introducing "expect" in makeTestGuess settings:
```
"guess" : helpers.makeTestGuess("America/Cancun", 
     { offset: false, abbr: true, expect: "America/Panama" }),
"guess" : helpers.makeTestGuess("America/Fort_Nelson", 
    { offset: true, abbr: true, expect: "America/Phoenix" }),
"guess" : helpers.makeTestGuess("America/Metlakatla", 
   { offset: true, abbr: true, expect: "America/Anchorage" }),
```
here it means that tz.guess() is expected to return "America/Panama" when timezone is set to "America/Cancun"


mj1856:
LGTM.  Thanks!

PR #608, opened 2018-05-04
ignore .idea files

jehy:
.idea files are getting published to npm (at least 05.26 has them inside), you should ignore them on publish.


ellenaua:
Thanks! This branch is conflicting (because of new release) with "develop" so I just made that change in "develop" and close this PR if you don't mind.

PR #610, opened 2018-05-04
ignore latest unpacked data on npm publish

jehy:
As far as I see, latest unpacked data is only used in grunt build task, which ignored in `.npmignore`.
So, having it in npm is pretty meaningless, and it is harmful because latest.json weights 2 MB.


ellenaua:
Thanks! This branch is conflicting (because of new release) with "develop" so I just made that change in "develop" and close this PR if you don't mind.

PR #613, opened 2018-05-07
Update to IANA tzDb 2018e

ellenaua:
What's done here: 
- updated to IANA 2018e database
- improved guess tests: changed `tasks/data-tests.js` so that it generates 2 guess tests for each timezone
- ensured that guess tests pass both for 2018e (latest) and 2018d databases (it means that database update did not break guess function, as it happened in previous release)


mj1856:
Don't forget to update the moment.js website.  It should be as simple as changing the dependencies in the project.json and committing.

Issue #614, opened 2018-05-09
Change year range for data file to 2013-2023


Schnark:
Currently, there is a file `moment-timezone-with-data-2012-2022.js` for those who want a smaller data file with only currently relevant data. For 2018 this range should be changed to 2013-2023, as it is supposed to be "current year +/-5 years" (previous change on this was https://github.com/moment/moment-timezone/commit/83e81ef127cba403aefef553c954d1ab17ef4c02).

aadityakulkarni:
Hi @Schnark I would like to take this up. I will create a PR for this.

aadityakulkarni:
Hi @Schnark I have created a PR for this issue. But the build failed. Can someone please help me fix this build issue?

andrewpanfelsherpa:
Referencing https://github.com/moment/moment-timezone/issues/363#issuecomment-235654248 , would it be possible to rename the file to something like "moment-timezone-with-data-truncated.js" so that the path doesn't need to continually change?

mj1856:
Thinking about this more, I believe the correct way to handle this is to do all of the following:

- Keep publishing the `2012-2022` file.  When we last changed from 2010-2020, a lot of folks were suddenly broken.  Don't want to repeat that.

- Start publishing multiple other files:
   - `2013-2023` - should have done this last year.  My bad.
   - `2014-2025` - this is where we should be now.
   - `PlusMinus5Years` - this should be always updated and floating.  Most users can take this one and the path won't continually change.
  - `1970-2030` - this range is much smaller than the "full" data, and seems to be prefered way to handle #697 based on comments there.

I'll pick this up for the next release, coming shortly.

mj1856:
Actually, I see no reason to publish older year range files.  We'll keep 2012-2022 intact, and add a 1970-2030 and a 5-year rolling file.

mj1856:
This is completed in version 0.5.24.  Thanks.

MatthewHerbst:
@mj1856 how does one use the new smaller file (1970-2030)?

andrewpanfelsherpa:
@MatthewHerbst For example, in my project:
```javascript
import Moment from "moment";
import MomentTZ from "moment-timezone/builds/moment-timezone-with-data-10-year-range";
```
Basically, the build contains the moment-timezone library _and_ the data for the date ranges, and can be substituted for the normal moment-timezone + data import.

More information can be found in the documentation: http://momentjs.com/timezone/docs/

[...]

PR #621, opened 2018-06-05
Changed year range for data file to 2013-2023

aadityakulkarni:
This will resolve issue #614 - Changed year range for data file to 2013-2023


mj1856:
Thanks for contributing, but I'll do this along with some of the other changes I mentioned https://github.com/moment/moment-timezone/issues/614#issuecomment-483785619

PR #622, opened 2018-06-06
Change moment to be a peerDependency and a devDependency

iMarv:
Currently, when using moment-timezone, moment is installed as a dependency.
This means if I have moment installed as a dependency in general, it won't be reused. Instead, moment-timezone uses its own version of moment.

This is leading to issues with a locale magically not being set or similar, as the "wrong" moment dependency is being used.

With this pull-request, moment is defined as a peer-dependency, which means it will be reused from an existing dependency. In a dev setup, we want to install moment in any case, this is why I added it to the devDependencies aswell.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,10 +26,11 @@
 		"url": "https://github.com/moment/moment-timezone/issues"
 	},
 	"license": "MIT",
-	"dependencies": {
+	"peerDependencies": {
 		"moment": ">= 2.9.0"
 	},
 	"devDependencies": {
+		"moment": ">= 2.9.0",
 		"grunt": "^1.0.1",
 		"grunt-contrib-clean": "^1.1.0",
 		"grunt-contrib-jshint": "^1.1.0",


dominykas:
This breaks things very subtly... It's a pity this is not v1 yet, so no way to indicate via semver and there's no tag under releases here (but thanks for the changelog!)

In our case, we never used `moment` directly and always relied on `moment-timezone` to bring it in (i.e. if you have `moment-timezone` - why would you need `moment` in your app?).

This means that since `npm` no longer breaks on missing peer deps for quite a while, you only get a warning (which is easy to miss) and your app breaks the moment (intended) you do `require('moment-timezone')` unless you have `moment` actually installed.

I would argue this should be unpublished and the earlier behavior restored and then possibly publish a v1 instead with this change, if it's really desirable?

Anyways, thanks for the good work, and I hope there won't be an entitled angry mob, which sometimes descends on maintainers when things break...

PR #645, opened 2018-06-30
Release 0.5.22

ellenaua:
1) Added `bin/moment-timezone.js` 
   which can be used to create `moment-timezone-with-data` for any range of years
   just run `moment-timezone build 2014-2024`

2) Improved how links are packed, multiple zones in one line
    "Australia/Brisbane|Australia/Lindeman|Australia/Queensland"


maggiepint:
New release needs to happen

PR #668, opened 2018-08-30
Fix for use when momentJS is nested.

darkenvy:
During the webpack process, momentJS is nested inside {default: ... }. A simple test asserts if moment is not in fact at the highest level, but nested one level deeper.

The if statement asserts that .version is missing while .default exists. A variable was made to avoid reassignment to a parameter (which is considered bad practice).

All tests pass. Works in my production software with this patch.

PS: This seems to be a problem others have as well: https://github.com/moment/moment-timezone/issues/449

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -15,7 +15,7 @@
 	} else {
 		factory(root.moment);                        // Browser
 	}
-}(this, function (moment) {
+}(this, function (_moment) {
 	"use strict";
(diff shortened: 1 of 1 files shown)


ellenaua:
Merged similar fix today https://github.com/moment/moment-timezone/pull/848/commits/0296382042a7b656f278c828eeaa38872a131e8c

PR #681, opened 2018-10-18
Update tzdata to 2018f

jacobwgillespie:
Updates the timezone data to the latest `2018f` via `grunt data`.


mj1856:
We'll just jump straight to 2018g, thanks.

PR #682, opened 2018-10-19
Phrasing "...out of, or in..."

Rinsington-DC:


mj1856:
We use the MIT license text verbatim as found here:
https://opensource.org/licenses/MIT

PR #689, opened 2018-10-28
Added 2018g timezones data set

alaouy:
Updates for the 2018g IANA timezones dataset


mj1856:
Thanks.  A few more steps needed to publish, but I got it from here.

PR #745, opened 2019-04-24
Timezone guess utc offset check

ccorcos:
Fixing an issue from https://github.com/moment/moment-timezone/issues/517.

Some background. CST stands for Central Standard Time and Chicago Standard Time. PST stands for Pacific Standard Time and Philippines Standard Time. This causes Intl API to be unreliable in older versions of Chrome and Electron.


ichernev:
Somehow this MR includes a bunch of generated files.

About this particular issue, it's best to let users decide whether to use Intl or not, check https://github.com/moment/moment-timezone/issues/873#issuecomment-1228473648

So I'd merge something that adds an options object to guess, backwards compatible with existing code, that lets users specify Intl or in-house or both. Then we could put a note about buggy Intl and let users decide how to handle it.

PR #777, opened 2019-09-05
Update moment to the latest version

jonathanchrisp:
## Overview
I need to use the latest version of moment within moment-timezone. 

When installing the current version of moment-timezone the version is locked down to version 2.22.2. Apologies for all the updates to the package-lock.json file, if you update this in a different manner please let me know and I can update.

In order to get around this problem I can install the latest version of moment within my project and import it with moment-timezone but, I would rather just install install moment-timezone and get the latest version as cleaner.


rafaelsales:
What's missing to merge this? A lot of apps are rendering the wrong time because the timezones are out of date.

PR #783, opened 2019-09-25
updating to use the latest 2019c data

ssherman:


ailtonbueno:
Hi @ssherman , congratulations.
The Brazil no longer has DST changes. The decree signed at 04/2019. I test this pull request code, but the time has changed to the old DST of the Brazil.

Can you check this new change to Brazilian time and include in this pull request?

PR #793, opened 2019-10-30
Update moment-timezone-with-data-10-year-range.js

kasa10:
+ 1 links


ichernev:
Sorry, we (the moment-timezone team) do the data bundling on release, and we follow IANA's releases (i.e don't invent or tweak the data). Not really sure what you changed in there, but a PR is not the way for it.

PR #839, opened 2020-05-08
Fixed changelog to represent proper release versions

Jackalex36:
The latest 2 versions stated in the current changelog were shifted, resulting in misleading versions (0.5.29 doesn't exist yet).


ellenaua:
Thanks @Jackalex36 !

Issue #871, opened 2020-07-10
.tz(.., true) incorrect around DST


mongoose700:
Fiddle: https://jsfiddle.net/r3qx7zh9/2/

```
const utc = moment.utc('2018-03-11T04:00');
const converted = utc.tz('America/Los_Angeles', true);
console.log(converted.toString());
console.log(converted.toISOString(true));

const direct = moment.tz('2018-03-11T04:00', 'America/Los_Angeles');
console.log(direct.toString());
console.log(direct.toISOString(true));
```

Output:
```
"Sun Mar 11 2018 05:00:00 GMT-0700"
"2018-03-11T04:00:00.000-08:00"
"Sun Mar 11 2018 04:00:00 GMT-0700"
"2018-03-11T04:00:00.000-07:00"
```

It's very strange that the converted moment has different offsets for `toString` and `toISOString`. It stops happening at '2018-03-11T10:00'.

mongoose700:
I've looked into the code a bit, and it seems that it's doing the conversion in two steps. First it converts the time from `2018-03-11T04:00Z` to `2018-03-10T20:00-08:00`, then it tries to move it into the future by eight hours, which puts it at `2018-03-11T04:00-08:00`, which is actually `2018-03-11T05:00-07:00` because of the DST shift.

ichernev:
Yep, I can verify that the bug exists. We could try to extract pieces (year, month, day, hour, minute, second) and re-create a new moment from that, it should work better.

ichernev:
Possibly related to #935

gilmoreorless:
Fixed by #1151

PR #872, opened 2020-07-13
Fix keeping time through tz to correctly handle being near DST boundaries

mongoose700:
This is a fix for https://github.com/moment/moment-timezone/issues/871

This change simplifies the process by just using `moment.tz(this.toArray(), name)`, so that it doesn't have to deal with offsets directly. I then also added logic to handle making it use the second occurrence of the time in the time zone if possible, if it was the second occurrence in the original time zone.

I'm not all that certain about the `var zone = mom._z || moment.defaultZone || getZone(guess());` check. It seems to match the logic for how a moment with an unspecified time zone behaves, but I haven't been able to confirm it.


ichernev:
@mongoose700 thanks for working on this!

Using `toArray()` seems like a very sensible thing to do. About adding all the additional code about repeated time, I'm less enthusiastic. 

About the guess() being injected in there, I'm even less enthusiastic, need to see how things are expected to work across the board.

PR #879, opened 2020-07-31
"Asia/Calcutta" added for ios

mahipalsingh-syt:


gilmoreorless:
The data used by Moment Timezone comes from the [IANA time zone database](https://www.iana.org/time-zones) and is [re-compiled for each database release](https://github.com/moment/moment-timezone/blob/develop/contributing.md#compiling-data). Therefore any changes to the `latest.json` file in a PR will be overwritten by the next release.

In this specific case, the `Asia/Calcutta` zone is a only link (alias) to the primary zone of `Asia/Kolkata` (after the city changed its official English spelling). This is why it doesn't appear in the `meta` data file, which only lists primary zones, not links.

You mentioned this is for iOS — is this related to the `moment.tz.guess()` API? I know there are some browsers that return `Asia/Calcutta` instead of `Asia/Kolkata` when asked for their user time zone.

ichernev:
Thank you @gilmoreorless for getting to the bottom of this.

Issue #798, opened 2019-11-07
Time is off by -1 hour on a 2099 date when switching from 0.5.25 to 0.5.27


bdavis-tc:
Simplified reproduction code:
https://runkit.com/bdavis-tc/5dc43803b63326001b4f6f53

Dates after year 2037 the 0.5.27 version misses the time by an hour.
My test date is in the year 2099.
Testing against timezone: EST (with Daylight Savings Time)

The 0.5.25 version seems to have accurate time all the way to year 2499.

Not sure if the 0.5.26 and 0.5.27 version introduced a bug, or just changed default behavior for how much data to include.

For now we will stick with 0.5.25, looking to improve my understanding and/or implementation to allow us to use the latest version again.

Thank you!

gilmoreorless:
This is probably due to the (accidentally?) reduced data size due to a change in release process from 0.5.26 onwards (#768).

micheleangioni:
Hi guys, I've lost around 3 hours today due to this :| 
One of our test dates is 2060 and since v0.5.26 the offset of dates with `Europe/Berlin` timezone is wrong. It seems to be indeed related to (#768).

andywu0408:
Following up on this - This problem still persists as of now for me.

eramitmittal:
Please look at #921 
Fork available at: https://github.com/eramitmittal/moment-timezone

ichernev:
Latest moment-timezone (0.5.37) has extended range. In the future we'll release on Linux :) check https://github.com/moment/moment-timezone/issues/999#issuecomment-1229214980

Issue #768, opened 2019-07-07
Release 0.5.26 has massively reduced the data range


gilmoreorless:
My moment-timezone webpack plugin has [a failing build](https://travis-ci.org/gilmoreorless/moment-timezone-data-webpack-plugin/jobs/555041551) when using version 0.5.26. After some investigation it seems the range of years in the compiled data has shrunk considerably from 0.5.25.

Compare the packed data files for [2019a](https://github.com/moment/moment-timezone/blob/develop/data/packed/2019a.json) and [2019b](https://github.com/moment/moment-timezone/blob/develop/data/packed/2019b.json). The file size has reduced by around 80%, and a quick switch between the two files shows that the count of abbreviations and offset has significantly changed for many zones.

Picking an example at random:

```
2019a.json: "America/Atikokan|LMT CST CDT CWT CPT EST|66.s 60 50 50 50 50|01212345|-32B5R.w UFdR.w 1in0 Rnb0 3je0 8x30 iw0|28e2",
2019b.json: "America/Atikokan|CST CDT CWT CPT EST|60 50 50 50 50|0101234|-25TQ0 1in0 Rnb0 3je0 8x30 iw0|28e2",
```

To see the impact of this, I wrote a quick script to log out the extremes of a particular zone, based on the packed data files:

```js
const moment = require('./moment-timezone-utils');
const versions = ['2018i', '2019a', '2019b'];
const name = 'Australia/Sydney';

function logDetails(version) {
  const data = require(`./data/packed/${version}.json`);
  const rawZone = data.zones.find(zone => zone.startsWith(`${name}|`));
  const { untils } = moment.tz.unpack(rawZone);

  const start = moment.utc(untils[0]);
  const end = moment.utc(untils[untils.length - 2]);

  console.log(`
Version: ${version}
Start:   ${start.toISOString()}
End:     ${end.toISOString()}`);
}

versions.forEach(logDetails);
```

The result of running this script in the root of the `moment-timezone` repo is as follows:

```
Version: 2018i
Start:   1895-01-31T13:55:08.000Z
End:     2499-10-03T16:00:00.000Z

Version: 2019a
Start:   1895-01-31T13:55:08.000Z
End:     2499-10-03T16:00:00.000Z

Version: 2019b
Start:   1916-12-31T14:01:00.000Z
End:     2037-10-03T16:00:00.000Z
```

[...]

Issue #825, opened 2020-03-22
Why Asia/Shanghai does not have LMT?


shellinger:
According to tzdb-2019c of IANA, Asia/Shanghai uses LMT before 1901, as file "asia" says:
\# Beijing time, used throughout China; represented by Shanghai.
Zone	Asia/Shanghai	8:05:43	-	LMT	1901
\t	\t	\t	8:00	Shang	C%sT	1949 May 28
\t	\t	\t	8:00	PRC	C%sT

But in moment-timezone-with-data.js, it uses the packed data as:
"Asia/Shanghai|CST CDT|-80 -90|010101010101010101010101010|-1c2w0 Rz0 11d0 1wL0 A10 8HX0 1G10 Tz0 1ip0 1jX0 1cN0 11b0 1oN0 aL0 1tU30 Rb0 1o10 11z0 1qN0 WL0 1qN0 11z0 1o10 11z0 1o10 11z0|23e6",

Apparently don't include the LMT part. How comes this inconsistence? Is it a bug?

BTW: What is the last number mean in the packed date (23e6 in above case). I can't find any explanation in all documents. Thanks!

gilmoreorless:
This looks like it's caused by a drop in data size in recent releases (see #768). Specifically, the latest data files are only handling 32-bit timestamps (which only allows for going back to `1901-12-13`).

Data files built in older releases contain the expanded range:

```bash
$ grep Shanghai data/packed/2019a.json | head -n1
"Asia/Shanghai|LMT CST CDT|-85.H -80 -90|0121212121212121212121212121|-2M0U5.H 1zWo5.H Rz0 11d0 1wL0 A10 8HX0 1G10 Tz0 1ip0 1jX0 1cN0 11b0 1oN0 aL0 1tU30 Rb0 1o10 11z0 1qN0 WL0 1qN0 11z0 1o10 11z0 1o10 11z0|23e6",
```

For your other question, the last value is the `population` field (in scientific notation, so `23e6` means `23,000,000`). This is only used for ranking zones when using `moment.tz.guess()`.

PR #881, opened 2020-08-01
When year is in some far future - use offsets of some similar year before 2038

ellenaua:
This is related to bugs:

https://github.com/moment/moment-timezone/issues/798
https://github.com/moment/moment-timezone/issues/768

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -169,6 +169,11 @@
 					return i;
 				}
 			}
+
+			var alternative = this._getAlternative(timestamp);
+			if (alternative) {
+				return this._index(alternative);
+			}
 		},
 
 		countries : function () {
@@ -200,8 +205,23 @@
 					return offsets[i];
 				}
 			}
+			
+			var alternative = this._getAlternative(timestamp);
+			if (alternative) {
+				return this.offset(alternative);
+			}
 
 			return offsets[max];
+		},
+		
+		_getAlternative: function (timestamp) {
+			var similarYear = findYearLike(new Date(timestamp).getFullYear());
+
+			if (similarYear) {
+				var result = new Date(timestamp);
+				result.setFullYear(similarYear);
+				return result.getTime();
+			}
 		},
 
 		abbr : function (mom) {
@@ -327,6 +347,16 @@
 			offset = offsets[i];
 			guesses[offset] = guesses[offset] || {};
 			guesses[offset][name] = true;
+		}
+	}
+
+	function findYearLike(year) {
+		for (var j = 2038; j > 2000; j--) {
+			if (new Date(year + '-01-01').getDay() === new Date(j + '-12-31').getDay() && 
+				new Date(year + '-01-01').getDay() === new Date(j + '-12-31').getDay()
+            ) {
+				return j;
+			}
 		}
 	}
 


moment-timezone.js
```diff
@@ -169,6 +169,11 @@
 					return i;
 				}
 			}
+
+			var alternative = this._getAlternative(timestamp);
```
mongoose700, line 173: I don't think it makes sense for this logic to be in `_index`, since it's not returning the index for an `until` that represents the timestamp. It would have the correct offset. I think it would be better to create a different method that wraps this information instead, and have the other method call the existing `_index`.

moment-timezone.js
```diff
@@ -330,6 +350,16 @@
 		}
 	}
 
+	function findYearLike(year) {
+		for (var j = 2038; j > 2000; j--) {
```
mongoose700, line 354: If we aren't worried about strange exceptions (like when a timezone skips a day entirely), we use the fact that there's a 28-year cycle in which the calendar repeats itself, instead of performing a loop.

moment-timezone.js
```diff
@@ -330,6 +350,16 @@
...
+			if (new Date(year + '-01-01').getDay() === new Date(j + '-12-31').getDay() &&
```
mongoose700, line 355: You're repeating the same check here instead of comparing the first day of the week of the target year with the last day of the week of the candidate year, instead of comparing the first and last days of the week of the target year with those of the candidate year.

moment-timezone.js
ellenaua: Yes, I already fixed that, will push tomorrow, it was beta version to check if that logics (of searching similar years) will work at all

moment-timezone.js
ellenaua: fixed that already, will push changes tomorrow

micheleangioni:
@ellenaua any news about this PR? :)

moment-timezone.js
micheleangioni: Any news about the fix @ellenaua ? :)

andywu0408:
Hi @ellenaua, May I ask if there is any update on this PR? Thanks a lot by the way!

eramitmittal:
Please have a look at #921

gilmoreorless:
The problem I raised in #768 is not fixed by this PR, or by #921. See #825 for a related issue in the other direction (very old timestamps).

The core problem is that the current build process is highly dependent on the version of `zdump` provided by the machine of the person building a release. Since 0.5.26, it looks like releases are being run on macOS, where the in-built `zdump` only produces 32-bit data files. Previously the releases included 64-bit data files (but I'm not sure which OS they were running on).

Ideally there'd be a standardised CI setup for building new data files (using GitHub Actions or similar), which could better guarantee data quality. I could take a look into doing this, but to be honest, I don't see many contributor PRs being merged on this repo, so I'm not sure it would be accepted.

ichernev:
> The core problem is that the current build process is highly dependent on the version of zdump

Ah, @gilmoreorless already troubleshooted this. Nice.

Closing this for now.

About using CI to release for a more stable result -- we could, as long as it's very transparent. Or we could have a small docker file in the repo and use that for zic/zdump.

PR #898, opened 2020-10-12
World Correction...

SCR-IR:


gilmoreorless:
The data used by Moment Timezone comes from the [IANA time zone database](https://www.iana.org/time-zones) and is [re-compiled for each database release](https://github.com/moment/moment-timezone/blob/develop/contributing.md#compiling-data). Any changes to the data will need to go through the tzdb mailing list first.

However, this specific change was already suggested on the mailing list a few months ago, and wasn't accepted. You can see the replies at https://mm.icann.org/pipermail/tz/2020-October/029408.html

PR #901, opened 2020-10-28
ZD_168565_Update_Casablanca

lmonterosa:
**DO NOT MERGE YET**
**DO NOT MERGE YET**
**DO NOT MERGE YET**

ZD ticket https://glghd.zendesk.com/agent/tickets/168565

Fixing Morocco timezone for Moment. Casablanca is supposed to be UTC +1 at this time. 

image


ichernev:
We take Zone data directly from IANA, so if IANA has released a new version we should just release an update. Ofc everybody can just compile the data and use any IANA data, regardless of whether we release or not.

If IANA has NOT released a new version you should urge them to, not us.

PR #912, opened 2020-11-14
Ability to use local abbreviations (many users request it)

ellenaua:
I added ability to use local abbreviations.
In 2017, IANA database deprecated these abbreviations but many users request it.

Usage example:
```
console.log(moment().tz('America/Sao_Paulo').format('HH:mm z'));  // 13:13 -03
 
moment.tz.useLocalAbbrs = true;
console.log(moment().tz('America/Sao_Paulo').format('HH:mm z'));   // 13:13 BRT
```

How it looks in packed.json:

"America/Sao_Paulo|LMT -03(BRT) -02(BRST)|36.s 30 20..."

@mj1856 what do you think about this?
By default it will return -03 as it does now.


gilmoreorless:
I know I'm not a maintainer, but I do follow the IANA updates closely. Personally, I don't think this is a good idea to implement, for a couple of reasons.

1. Correctness. The IANA database maintainers didn't just deprecate these specific abbreviations. They were removed entirely, because **they had been completely made up by the maintainers** and weren't official labels used by the relevant countries. Yes, they had been in use for a long time, and some of them had become common, but that doesn't stop them from being incorrect.
2. Maintenance. The JSON file containing the old abbreviations looks to be hand-coded, or generated by a script that isn't part of the PR. When (not if) the tzdb removes more abbreviations, how will that file be updated as part of the data update and release process?

I know some users are confused about why the abbreviations have been removed, but documentation and education can be provided. Given the maintenance-only status of this project, I think sticking as closely as possible to the correct IANA source data is more important.

Also, this PR is a bit confusing in its structure. There are 3 separate commits with identical messages, and a bunch of other completely unrelated commits relating to releases. That makes it much harder for reviewers to identify what the relevant changes are.

mj1856:
I tend to agree with Gilmore's points.  Thanks for your efforts, but I recommend against this.

ramseth001:
@ellenaua @mattjohnsonpint As a developer, I understand the valid points put forward by @gilmoreorless. But, where do we go from here ? How do we display the time of timezones which do not have abbreviations to people who do not understand the concept of IANA?

ichernev:
@ramseth001 this question is more a matter of pleasing a particular set of users accustomed to a particular timezone abbreviation.

If you have a project that displays timestamps in different timezones and have people complain that they don't know what -03 means and want XYZ instead, just hardcode in your own code base that people in a particular zone prefer particular abbreviations. Just don't make this OUR job, because, as @gilmoreorless pointed out, it's not official and it's not correct.

PR #920, opened 2020-12-23
fixing build on non mac systems like linux

eramitmittal:
When TZ data is updated on linux using grunt data command, zdump output in case of no transition is not to the linking of grunt data-collect task. This fix ensures that usable zdump files get generated even in such cases

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,8 @@
+### Moment-timezone version which you use:
+
+Version:
+
+_Note: many issues are resolved if you just upgrade to the latest version_
+
+### Issue description:
+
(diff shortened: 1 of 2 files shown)


ichernev:
Can you give an example timezone that falls in this? Currently there is [this](https://github.com/moment/moment-timezone/blob/develop/tasks/data-collect.js#L31) inside collect, but what happens on length 13 is the question ;-) I also added the failed check (previous line).

Now the grunt task works well in the last few IANA releases.

gilmoreorless:
There were some cross-platform issues with the output of `zdump`. This solved one of them, but not all. I've done a wider sweep of the `zdump`-related code in #1059, which should fix this issue as well.

PR #921, opened 2020-12-23
IANA TZ database 2020d but with data beyond 2038

eramitmittal:
I tried re-downloading IANA TZ DB 2020d and re-building the moment-timezone library on Mac and I faced the issue of TZ data beyond 2038 not getting captured.
Next I tried the same steps on linux (after fixing data-zdump task, details in #920 ) and this time the built files had data beyond 2038, in some cases till 2499 end of year.

I performed the following steps

1. Ran grunt data:2020d on linux
2. Replaced latest.json with 2020d.json in pack and unpack folders (did this as curent IANA TZ DB version is 2020e, I wanted to keep the DB at same version as currently released one)
3. Changed version to 0.5.33 in package,json and at other places
4. Ran grunt build
5. Ran grunt uglify
6. Ran grunt nodeunit


eramitmittal:
While waiting for this request to get merged, fork available at https://github.com/eramitmittal/moment-timezone

gilmoreorless:
@eramitmittal This is useful for fixing #768! But as far as I can see, there are no updates to the build process. This means that the next time a release is built by one of the maintainers, the data will be back to the 2038 limit again, depending on the machine it's built from.

SanthoshYAR:
> While waiting for this request to get merged, fork available at https://github.com/eramitmittal/moment-timezone

Time is off by -1 hour on a 2099 date when switching from 0.5.25 to 0.5.27 #798
This will happen for dates where starts from 2038, can you please check and confirm

ichernev:
The reason for the "reduced range" is macos vs linux release machine. Unfortunately build process depends on zic/zdump, which come from OS. Check https://github.com/moment/moment-timezone/issues/999

PR #936, opened 2021-04-06
zonesForCountry to return null for non-existant countries

alexanderweiss:
`zonesForCountry` will return `null` if the country can't be found found: https://github.com/moment/moment-timezone/blob/ffc41900fed9efa9a31e87f0eadd66ec86b5046a/moment-timezone.js#L511


ichernev:
For some reason this MR is messed up too, not sure why.

What is the reason for this, to figure out if a country is valid or not? I'm not so sure we should change the API, it could break people's code. It's better to add a new API to check if a country exists, but then this is kinda outside the scope of moment-timezone. If a country doesn't have timezones it's a pretty clear indicator something's off :)

gilmoreorless:
Closing as stale.

alexanderweiss:
This PR simply added `null` as a possible return type, as moment-timezone can actually return `null` from this function (as linked to code)

PR #944, opened 2021-06-17
added US/Pacific-New zone

kirankumar2865:
since it's used by so many customers and moment-timezone can't convert it.


max-fex (approved):

gilmoreorless:
The data used by Moment Timezone comes from the [IANA time zone database](https://www.iana.org/time-zones) and is [re-compiled for each database release](https://github.com/moment/moment-timezone/blob/develop/contributing.md#compiling-data). The `US/Pacific-New` identifier was a long-deprecated link that was [removed from the tzdb in release `2018a`](http://mm.icann.org/pipermail/tz-announce/2018-January/000048.html):

> The default installation procedure no longer creates the
     backward-compatibility link US/Pacific-New, which causes
     confusion during user setup (e.g., see Debian bug 815200).
     Use 'make BACKWARD="backward pacificnew"' to create the link
     anyway, for now.  Eventually we plan to remove the link entirely.

If you have many customers using that identifier, they should be encouraged (or forced) to use something else. `US/Pacific-New` was intended to represent daylight saving rules that were never actually made into law ([reference](http://catless.ncl.ac.uk/Risks/13.87.html#subj1)).

Additionally, this PR only updates `data/packed/latest.json`. This is a file that's auto-generated by the build process. So the next data release will completely erase this change.

ichernev:
Yep, no-go, thanks @gilmoreorless

PR #953, opened 2021-09-28
Don't try to guess timezone for UTC

### LINKED ISSUE #559: Incorrect Timezone in IE 8
On a Windows 10 laptop in IE 11 when running `moment.tz.guess()` it returns "Africa/Abidjan" when the timezone is set to "(UTC) Co-ordinated Universal Time". On the same laptop in Chrome it correctly returns "UTC"

Running the following code produces the following outputs:

```
console.log((new Date()).toString());
console.log((new Date()).toLocaleString());
console.log((new Date()).getTimezoneOffset());
console.log(navigator.userAgent);
console.log(moment.version);
```

**Internet Explorer 11 (London)**
undefined
Fri Nov 10 2017 10:37:23 GMT+0000 (GMT Standard Time)
10/11/2017 10:37:23
0
Mozilla/5.0 (Windows NT 10.0; WOW64; Trident/7.0; .NET4.0C; .NET4.0E; rv:11.0) like Gecko
2.19.1

**Edge (London)**
Fri Nov 10 2017 10:39:42 GMT+0000 (GMT Standard Time)
10/11/2017 10:39:42
0
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36 Edge/16.16299
2.19.1
undefined

**Internet Explorer 11 (UTC)**
undefined
Fri Nov 10 2017 10:42:32 GMT+0000 (Co-ordinated Universal Time)
10/11/2017 10:42:32
0
Mozilla/5.0 (Windows NT 10.0; WOW64; Trident/7.0; .NET4.0C; .NET4.0E; rv:11.0) like Gecko
2.19.1

**Edge (UTC)**
Fri Nov 10 2017 10:44:22 GMT+0000 (Co-ordinated Universal Time)
10/11/2017 10:44:22
0
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36 Edge/16.16299
2.19.1
undefined

bk2204:
In the IANA database, there are two main kinds of time zones.  First, there are the timezones, such as America/Los_Angeles, that reflect a physical location.  When a user specifies one of this type of time zone, that is usually because they are or were physically present in that time zone, or something relevant to them is located in that physical time zone.

The other kind, like UTC and the GMT offset time zones, represents some arbitrary (possibly zero) offset from UTC or another time reference. These usually do not indicate the user is physically present, and are used when the user explicitly prefers a fixed offset.  This is common for server infrastructure which must be maintained across time zones.

The code which checks for time zones, however, looks to see if the user's specified time zone is greater than three characters.  If it is not, it attempts to guess the time zone by rewriting the offset into some time zone in that offset.  This unfortunately rewrites "UTC" into the first alphabetical entry which represents it, which is "Africa/Abidjan".

As mentioned above, while it is the case that Abidjan (and, for that matter, all of Côte d'Ivoire) are in the same offset as UTC, the two time zones provide semantically different meanings.  To avoid this problem, let's remove the restriction on length to allow users to specify a value such as "UTC" or "GMT" if that's the best option, and simply check that the length of the name is nonzero instead.

Fixes #559

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -359,7 +359,7 @@
 		// use Intl API when available and returning valid time zone
 		try {
 			var intlName = Intl.DateTimeFormat().resolvedOptions().timeZone;
-			if (intlName && intlName.length > 3) {
+			if (intlName && intlName.length) {
 				var name = names[normalizeName(intlName)];
 				if (name) {
 					return name;


bk2204:
Uh, no, thanks.  I don't sign CLAs.  Never mind, then.

PR #955, opened 2021-10-09
Allowed get packed data from zone

surfdude75:
This is usefull on server side applications to enabled send unique timezone data to lightweight clients.


ichernev:
@surfdude75 I'm completely lost in this PR, can you drop all the generated code and explain what you're trying to achieve?

surfdude75:
I just added the packed data accessible in Zone element.
I use it So I can sent it from server side application a lightweight version of moment I created for thin clients.
Here is the thin client I use https://github.com/surfdude75/moment-timezone-micro
Did you get the point?

gilmoreorless:
The CLA hasn't been signed, and this PR has many conflicts. Additionally, I think a server-side application that needs the packed data can just read the `data/packed/latest.json` file directly, rather than adding it to the public API.

surfdude75:
The point was to be able to provide the data to a client side app that need this timezone data on demand.

PR #963, opened 2021-11-09
2021e

ellenaua:


gilmoreorless:
@ellenaua Can you please not force-push changes to the `develop` branch? This repository has hundreds of forks that use `develop` as the default branch. Every forced push to the default branch breaks those forks.

PR #973, opened 2022-02-17
Do not fallback to now on invalid moments

lvhuiru520:
when moment object is invalid, should not get the current time


ichernev:
@lvhuiru520 can you please also supply unit tests to showcase what exactly should and shouldn't work. Also, how are you supposed to get current-time after the patch.

gilmoreorless:
Closing as stale, since the feedback questions weren't answered.

PR #976, opened 2022-03-05
 add package: "adjust-timezone "

vipulvyas:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
 			"integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
 			"dev": true,
 			"requires": {
-				"@babel/highlight": "7.0.0"
+				"@babel/highlight": "^7.0.0"
 			}
 		},
 		"@babel/generator": {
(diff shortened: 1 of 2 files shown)


gilmoreorless:
How does this package improve Moment Timezone? You've added it to the `dependencies` list but not added any code to call it.

I'm assuming there's a good intention behind this PR. But at face value it's looking like just self-promotion, without any benefit to this project.

vipulvyas:
First of all thank you for responding it and I am new to open source community let me current if i am thinking in wrong direction.
we can say `adjust-timezone` is adding new feature in `moment-timezone` which solve problem of passing date-time in different timezone. 
Suppose our client is in different region than server region. so whenever we pass date-time to client to server or server to client then and if we do any process on that date-time then it will automatically converted to that timezone and we can not get correct date-time.`adjust-timezone` will convert date-time into UTC with the help of time offset. Because of UTC is standard at all places. so it will give us correct date-time. 
**`adjust-timezone` is not improving existing functionality but it is adding feature.**

gilmoreorless:
Thanks for the explanation. Moment Timezone already has some features to convert between server and client times, if the time zones of each machine are known. https://momentjs.com/timezone/docs/#/using-timezones/converting-to-zone/

Right now this PR isn't adding any feature to Moment Timezone itself, because there are no changes to the Moment API at all. If users want to use `adjust-timezone` alongside `moment-timezone`, I think it's best for them to install the packages as separate dependencies. I'm not one of the project maintainers, but I don't think this PR will be accepted in its current form.

vipulvyas:
Yes, `moment timezone` already has feature to convert between server to client time, but adjust-timezone help in machine timezone which we don't know.
But i got your point we can use `adjust-timezone` alongside `moment-timezone` to installing as separate dependencies.
Thanks :)

ichernev:
Yeah, this thing is a no-go. People should strive to make composable libraries, not compose them into one big unmaintainable mess. If moment-timezone users feel like they need `adjust-time`, fine, but we're not pre-bundling it.

PR #996, opened 2022-08-15
Update to timezone database 2022b

repl-angrez-singh:


nickosepulveda (commented):
In Chile, the time change is approaching very soon. We are waiting for the new version to install it.

moranza:
@nickosepulveda Can the pr be merged? We would like to have the 0.5.35 version before 4.9 :)

jasonmobley:
@ellenaua we could use your help getting this change merged and a new release published ASAP due to the DST policy change in Chile mentioned above by others.

https://www.timeanddate.com/news/time/chile-dst-delay-2022.html

joebehymer:
Hello, we would very much appreciate this getting merged as soon as possible.

We are considering whether to copy this branch into a local fork of moment-timezone to work around this.  Getting this merged will save us quite some time.  Cheers.

lhorie:
Has this been tested? With `"moment-timezone": "repl-angrez-singh/moment-timezone" in package.json, and the following test script

```
const moment = require('moment-timezone');

console.log(moment('2022-08-06T12:00:00Z').tz('Europe/Athens').format('ha z'))
console.log(moment('2022-08-06T12:00:00Z').tz('Europe/Kiev').format('ha z'))
console.log(moment('2022-08-06T12:00:00Z').tz('Europe/Kyiv').format('ha z'))
```

I get the following console output

```
3pm EEST
3pm EEST
Moment Timezone has no data for Europe/Kyiv. See http://momentjs.com/timezone/docs/#/data-loading/.
5am 
```

The expected output is `3pm EEST` x 3

lhorie:
@alexmaie I went ahead and published a fork including 2022c here https://www.npmjs.com/package/moment-timezone-fork

alexmaie:
@lhorie  thanks for the update !!!

ichernev:
Hey, 0.5.36 is now released (featuring 2022c tz data).

@nickosepulveda the way moment-timezone works is that you have code and you have data, and you can always provide your own (newer) data if you'd like. So next time, instead of waiting for somebody on the internet to fix things in your codebase, you can compile any data you want with `grunt data:2022c` (for example) and then put `data/packed/latest.json` somewhere accessible on your server, then fetch it, parse it and pass it to `moment.tz.load(dataObject)`.

There is also little need to provide a fork just for an updated data file, but ... whatever floats your boat I guess.

joebehymer:
Thank you @ichernev 

Maybe we can all take this as a wake up call to migrate away from moment :)

lhorie:
@ichernev For context, we maintain a large monorepo w/ hundreds of services owned by different teams, so loading data files via code changes is not really feasible, especially on short notice.

If using `grunt data` + .tz.load() is the recommendation (even after considering that moment-timezone is used in both server and browser contexts, and considering bundle size concerns), it might be a good idea to document that.

PR #1000, opened 2022-08-31
fix(npm) add compressed archives to ignore

NewEraCracker:
image

Looks like new version has some junk published by mistake while old version did not.


gilmoreorless (approved):
Nice and simple. Thanks!

PR #1002, opened 2022-09-11
Migrate CI from Travis to GitHub Actions

### LINKED ISSUE #721: Travis CI build failing
Locally `grunt release` succeed after IANA data update but on Travic it continue to fail? Why?
The pull request is https://github.com/moment/moment-timezone/pull/720

gilmoreorless:
Since `travis-ci.org` is no longer working, this PR migrates the tests to use GH Actions instead.

The main thing to get feedback on is which versions of Node.js to run tests for. I've copied the technique used by [the TypeScript repo](https://github.com/microsoft/TypeScript/blob/12ab0fea9f45c48ea644c8273bedde235c8bcc61/.github/workflows/ci.yml) by having a rolling set of versions:

1. Latest, bleeding-edge (currently `18.x`)
2. Most recent LTS (currently `16.x`)
3. Previous LTS (currently `14.x`)
4. **Edit:** Previous-again LTS (currently `12.x` — officially end-of-life but likely still used in some production environments)

This means less maintenance of the builds long term, but at the cost of not explicitly testing older versions.

Fixes #721


ichernev:
Looks great!

PR #1009, opened 2022-10-14
data: Add 2022e

gilmoreorless:
Contains the following changes from tzdb:

```
Release 2022e - 2022-10-11 11:13:02 -0700

  Briefly:
    Jordan and Syria switch from +02/+03 with DST to year-round +03.

  Changes to future timestamps

    Jordan and Syria are abandoning the DST regime and are changing to
    permanent +03, so they will not fall back from +03 to +02 on
    2022-10-28.  (Thanks to Steffen Thorsen and Issam Al-Zuwairi.)

  Changes to past timestamps

    On 1922-01-01 Tijuana adopted standard time at 00:00, not 01:00.

  Changes to past time zone abbreviations and DST flags

    The temporary advancement of clocks in central Mexico in summer
    1931 is now treated as daylight saving time, instead of as two
    changes to standard time.


Release 2022d - 2022-09-23 12:02:57 -0700

  Briefly:
    Palestine transitions are now Saturdays at 02:00.
    Simplify three Ukraine zones into one.

  Changes to future timestamps

    Palestine now springs forward and falls back at 02:00 on the
    first Saturday on or after March 24 and October 24, respectively.
    This means 2022 falls back 10-29 at 02:00, not 10-28 at 01:00.
    (Thanks to Heba Hamad.)

  Changes to past timestamps

    Simplify three Ukraine zones to one, since the post-1970
    differences seem to have been imaginary.  Move Europe/Uzhgorod and
    Europe/Zaporozhye to 'backzone'; backward-compatibility links
    still work, albeit with different timestamps before October 1991.
```


ichernev (approved):
LGTM

gilmoreorless:
> Next time you can put the country change separately (or we should just automate the file as well...).

Good call. When I next get a good chunk of time I'm going to experiment with an automated workflow for data updates. Ideally we could just click a button in GitHub to generate the latest data, without relying on having exactly the right versions of things on our local machines.

PR #1057, opened 2023-04-25
Include isdsts field in data

ethanshar:
isdsts field is missing from data 
This PR adds it to the relevant data tasks


gilmoreorless:
Thanks for the PR, but I'm a little confused about its purpose.

Moment Timezone doesn't add the `isdst` field to the data files because the library doesn't use that field for calculations. There are no changes to the calculation code here that would require the data. Could you elaborate about why this is needed?

Also, this PR only adds the field to the `unpacked` data file, but not the `packed` file that is actually shipped with the library.

Issue #1075, opened 2023-10-12
moment().valueOf returns a value even though it's invalid


tmkx:
### Environment

```js
moment.tz.setDefault();
t.ok(
  Number.isNaN(moment("2023/10/06", ["H", "HH", "HH:mm"], true).valueOf()),
);

moment.tz.setDefault('America/New_York');
// not expected
t.ok(
  Number.isNaN(moment("2023/10/06", ["H", "HH", "HH:mm"], true).valueOf()),
);
```

### Issue description

when defaultZone is specified, `valueOf` will return a value even though it's invalid.

it seems to be caused by:
https://github.com/moment/moment-timezone/blob/dffed7a8a9936e5be023ca76d389c258f1746dcd/moment-timezone.js#L643-L646

a fix that works:
```diff
- if (zone && needsOffset(mom) && !mom._isUTC) { 
+ if (zone && needsOffset(mom) && !mom._isUTC && mom.isValid()) { 
    mom._d = moment.utc(mom._a)._d;
    mom.utc().add(zone.parse(mom), 'minutes');
  }
```

sslincoco:
您的邮件已收到,我会尽快给您回复。

gilmoreorless:
Thanks for the detailed report. I've merged #1082 which contains the fix, and added your example as a test case in [ece926a](https://github.com/moment/moment-timezone/commit/ece926a59f53c72b51e8250ef97393900c0b7227).

PR #1082, opened 2023-12-17
Patch-1075

raphyluke:
Issue #1075 as been patched.
Added in the bugged code a condition to check if it's valid.


gilmoreorless:
Thanks!

PR #1084, opened 2023-12-26
data: Add 2023d

tmkx:
close #1083

PR #1085, opened 2023-12-27
data: Add 2023d

### LINKED ISSUE #1083: 2023d is out
### Issue description

Timezone database 2023d (Released 2023-12-21) is released

https://www.iana.org/time-zones

gilmoreorless:
This is a rework of #1084 using a freshly-compiled `zic` and `zdump` from the IANA source. There turned out to be two separate bugs at play with the initial data build:

1. macOS's `zdump` was previously very old and only handled 32-bit data, which this project's build process could catch and warn about. The latest Sonoma releases have updated to `zdump` from `2022g` which works _just enough_ to look like it's fine.
2. `zic` and `zdump` had bugs that inadvertently dropped some transitions from a small number of zones. This was fixed in `2023d`, which is why zones like `America/Ciudad_Juarez` now have more data without the source files changing. (It only affects transitions after the year 2400, so not a big deal.)
3. This lead to the combination of `zdump@2022g` compiling `data@2023d` on macOS, which look like it worked properly, but was subtly wrong. Zones like `America/Godthab` dropped any transitions after 2023, which was definitely not correct. Manually compiling `zdump` from the `2023d` source fixed that issue in this PR.

Interestingly, the new (previously-missing) data for `America/Ciudad_Juarez` also contains some buggy transitions right after the previous data cut-off. There's possibly another bug lurking in tzcode somewhere.

---

The PR contains the following changes from tzdb (edited to remove details irrelevant to this project):

```
Release 2023d - 2023-12-21 20:02:24 -0800

  Changes to future timestamps

    Ittoqqortoormiit, Greenland (America/Scoresbysund) joins most of
    the rest of Greenland's timekeeping practice on 2024-03-31, by
    changing its time zone from -01/+00 to -02/-01 at the same moment
    as the spring-forward transition.  Its clocks will therefore not
    spring forward as previously scheduled.  The time zone change
    reverts to its common practice before 1981.

    Fix predictions for DST transitions in Palestine in 2072-2075,
    correcting a typo introduced in 2023a.

  Changes to past and future timestamps

    Vostok, Antarctica changed to +05 on 2023-12-18.  It had been at
    +07 (not +06) for years.  (Thanks to Zakhary V. Akulov.)

    Change data for Casey, Antarctica to agree with timeanddate.com,
    by adding five time zone changes since 2020.  Casey is now at +08
    instead of +11.

  Changes to past tm_isdst flags

    Much of Greenland, represented by America/Nuuk, changed its
    standard time from -03 to -02 on 2023-03-25, not on 2023-10-28.
    This does not affect UTC offsets, only the tm_isdst flag.
    (Thanks to Thomas M. Steenholdt.)

  Changes to code

    zic no longer mishandles data for Palestine after the year 2075.
    Previously, it incorrectly omitted post-2075 transitions that are
    predicted for just before and just after Ramadan.  (Thanks to Ken
    Murchison for debugging help.)
```

Closes #1083


gilmoreorless:
I've sent a message to the tzdb mailing list to check about the strange `America/Ciudad_Juarez` data: https://mm.icann.org/pipermail/tz/2023-December/033392.html

PR #1097, opened 2024-02-08
reset to defaultZone

gronke:
Allow resetting to defaultZone by calling `tz(null)`.

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -598,7 +598,7 @@
 			out.add(zone.parse(out), 'minutes');
 		}
 
-		out.tz(name);
+		out.tz(name ?? undefined);
 
 		return out;
 	}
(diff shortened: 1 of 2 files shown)


gilmoreorless:
Hi, thanks for your contribution. Unfortunately this PR won't be accepted into the project, for a few reasons:

* This is technically a breaking API change. Right now, _any_ falsey value passed into `.tz()` is treated the same as not providing an argument. This includes `false`, `undefined`, `null`, `''` and `0`. (I'm not saying this is an ideal API, just the reality of the situation.) This PR would change the behaviour of just one of those cases, unexpectedly.
* The project is in maintenance mode, and the [contributing guidelines](https://github.com/moment/moment-timezone/blob/develop/contributing.md) explicitly say "new features or API changes are **unlikely to be accepted** as pull requests."
* The new code is adding very modern ES syntax into an ES5 codebase that has no transpilation step.
* There's a fairly simple workaround for resetting a instance's zone, if you know that there is a default zone set up: `instance.tz(moment.defaultZone.name)`

Issue #1066, opened 2023-05-27
Czechia and Slovakia timezone mixup


nexpid:
### Environment

in **data/packed/latest.json**, timezones **Europe/Prague** and **Europe/Bratislava** are marked as the same timezone:

https://github.com/moment/moment-timezone/blob/f7d8fc2d42a832b390c3f770ec5d52b4803a1a14/data/packed/latest.json#L572

country code **SK** includes **Europe/Prague** (incorrect) and **Europe/Bratislava** (correct):

https://github.com/moment/moment-timezone/blob/f7d8fc2d42a832b390c3f770ec5d52b4803a1a14/data/packed/latest.json#L803

### Issue description

As mentioned above, country code **SK** also includes **Europe/Prague**, even though Prague is in Czechia, not Slovakia.
![image](https://github.com/moment/moment-timezone/assets/60316309/1bce6a7f-4bb3-4281-a788-d6d82037a16d)

Running this code in the latest version of moment-timezone also yields wrong results:
```js
moment.tz.zonesForCountry("CZ"); // => [ "Europe/Prague" ] (✅)

moment.tz.zonesForCountry("SK"); // => [ "Europe/Bratislava", "Europe/Prague" ] (❌)
```

gilmoreorless:
This is a similar issue to #1005, and I'll tweak my comment from there.

The data comes from the IANA time zone database (tzdb). There have been some changes with recent releases of tzdb, where time zones that have the same data since 1970 have been merged together, even if they're in different countries. This has definitely caused some confusion here and in other software projects.

The important thing is that identifiers like `Europe/Prague` are **labels** for regions of the world where the clocks agree. The country data you've quoted is not saying that Prague is in Slovakia. It's saying that the time in Slovakia can be represented by a time zone that happens to have the label `Europe/Prague`.

This can be seen in the tzdb source [`zone1970.tab`](https://data.iana.org/time-zones/data/zone1970.tab):

> CZ,SK	+5005+01426	Europe/Prague

That line says that Czechia and Slovakia can both be represented by the same time zone rules. This would be equally true if the labels didn't have city names at all, but instead were named something like `Europe/Zone073`.

Note that `Europe/Bratislava` doesn't appear in that `zone1970.tab` file, because it's just a backwards-compatibility link to `Europe/Prague`. Moment Timezone adds the compatibility links to the country lists as well, based on the older (and deprecated) [`zone.tab`](https://data.iana.org/time-zones/data/zone.tab).

[...]

PR #1098, opened 2024-02-13
Update 2024a.json

wiserweb:
Removed Phoenix, Puerto_Rico, Panama which are not located in Canada and do not function as timezones for the country's zone.

### DIFF
--- a/data/meta/2024a.json
+++ b/data/meta/2024a.json
@@ -341,9 +341,6 @@
 				"America/Whitehorse",
 				"America/Dawson",
 				"America/Vancouver",
-				"America/Panama",
-				"America/Puerto_Rico",
-				"America/Phoenix",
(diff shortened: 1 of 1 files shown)


gilmoreorless:
Hi, thanks for your contribution, but I'm going to close this PR for two reasons:

1. The data files are auto-generated from the IANA tzdb releases. Any manual changes like this will be overwritten the next time a tzdb release happens. (Additionally, most installations of Moment Timezone read from a duplicate `latest` file that hasn't been altered in this PR.)
2. I know it's confusing and not obvious, but the data file is correct _for what it's intended to track_. The short version is "it's not saying that all these zones are definitely in Canada, just that time in Canada _can_ be represented by these zones". The longer version is https://github.com/moment/moment-timezone/issues/1066#issuecomment-1574578959

PR #1112, opened 2024-07-09
add tracking to the timezone error

fouadgcm:


gilmoreorless:
I see this PR is already closed, but for future reference, I will reject any PRs that don't provide a description. You need to put in at least a tiny bit of effort to explain to a project's maintainers why they should accept your proposed changes.

PR #1120, opened 2024-10-03
Add 2024b

### LINKED ISSUE #1119: 2024b
Hi there, any ETA on when 2024b will be available? Thanks.

er0k:
I realize contributor PRs to update the data are generally not accepted, but here is one anyway, just in case 🙂 

closes #1119


gilmoreorless:
Thanks for the prompt for me to fix the automated builds. 😄 

The reason we don't currently accept contributor PRs for data updates isn't anything personal, it's simply about having a standardised build environment. The data build process (for better or worse) relies on the binaries `zic` and `zdump`, which can vary wildly between platforms. Our CI process actually builds those binaries from scratch using the source code of the matching IANA release, guaranteeing compatibility.

As an example, this PR is very close to the automated #1121, but there are some subtle bugs that were introduced (presumably by an _ever so slightly_ out-of-date `zic` and/or `zdump`). In this PR, the zones `America/Godthab`, `America/Nuuk`, `America/Scoresbysund`, `Asia/Gaza` and `Asia/Hebron` contain data changes that shouldn't be there. This is a problem that I've seen only on recent versions of macOS, due to the specific versions of the provided binaries (see #1085 for a full description).

Issue #1017, opened 2022-11-28
Incorrect value when adding week to date and crossing PST daylight savings 


fpintos:
### Moment-timezone version which you use:
Version: "moment-timezone": "0.5.39" and "0.5.21"

### Issue description:
Hello,

I want to confirm if I'm using something incorrectly or if there is an issue:

Adding a week to 2018-03-04T02:00:00-08:00 in PST gives me different results in moment, browser's Date and dotnet.

For example, with 5.39:
```js
var moment = require("moment-timezone");
console.log(
  moment.utc("2018-03-04T02:00:00-08:00").tz("America/Los_Angeles").add(1, "weeks").toISOString(true)
);

OUTPUTS "2018-03-11T02:00:00.000-07:00"  <- Notice the "2AM" and "-7:00".
```

With 5.21:
```
OUTPUTS: "2018-03-11T01:00:00.000-08:00"  <- Notice the "1AM" and "-8:00".
```

However, browsers in a machine running PST, give a different date if we add 7-days' worth of milliseconds:
```js
new Date(new Date("2018-03-04T02:00:00-08:00").getTime() + 604800000)
OUTPUTS: "Sun Mar 11 2018 03:00:00 GMT-0700 (Pacific Daylight Time)"  <- Notice the "3AM" and "-7:00"
```

dotnet seems to agree with the 3AM value (in a machine running PST):
```cs
TimeZoneInfo.ConvertTime(DateTimeOffset.Parse("2018-03-04T02:00:00-08:00").AddDays(7), TimeZoneInfo.Local).ToString("o")
OUTPUTS: "2018-03-11T03:00:00.0000000-07:00"  <- Notice the "3AM" and "-7:00"
```

Two questions:
- why values returned by moment changed between 5.21 and 5.39? change in timezone data?
- why values returned by moment do not match what browsers and dotnet return?

Thanks

gilmoreorless:
> why values returned by moment changed between 5.21 and 5.39?

The earlier behaviour in `0.5.21` was a bug (#738) where the old offset wasn't cleared correctly around time zone transitions. This was fixed in version `0.5.24`.

> why values returned by moment do not match what browsers and dotnet return?

This looks like a different bug in Moment Timezone, while the dotnet behaviour is correct.

1. Adding 1 week (or 7 days) to `2 AM, March 4 2018` for Los Angeles produces `2 AM, March 11 2018`.
[...]

PR #1124, opened 2024-12-05
fix: update offset error with daylight saving time

shaodowm:
fix the dst offset calculate error of https://github.com/moment/moment-timezone/issues/1074 and https://github.com/moment/moment-timezone/issues/1017. When update the offset, the hole time (ignore time) of dst should not keepTime in the moment. Here is one way to find the hole time in function updateOffset. Please help to check if any problem.

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -652,7 +652,15 @@
 				offset = offset / 60;
 			}
 			if (mom.utcOffset !== undefined) {
+				var _offset = mom._offset;
 				var z = mom._z;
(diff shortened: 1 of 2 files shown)


gilmoreorless:
Thanks for your contribution and sorry for the delayed response. After a deep review of open bugs and PRs, we've found that this PR does fix some of the issues, but missed several cases relating to other ways of setting the time.

Closing this in favour of #1151 which provides a more comprehensive fix.

Issue #867, opened 2020-06-19
Incorrect result when parsing time in a timezone


createthis:
# Repro steps:

```js
import moment from 'moment-timezone';
import tk from 'timekeeper';
tk.freeze(moment('2020-06-19T20:00:00-04:00').toDate());
console.log('now=',moment().format());
const eight_pm_today = moment.tz('8:00 PM', 'h:mm A', 'America/New_York');
console.log('eight_pm_today=',eight_pm_today.format());
```

# Result:
```
now= 2020-06-19T20:00:00-04:00
eight_pm_today= 2020-06-20T20:00:00-04:00
```

# Expected Result:
```
now= 2020-06-19T20:00:00-04:00
eight_pm_today= 2020-06-19T20:00:00-04:00
```

(eight_pm_today is showing as the 20th rather than the 19th)

ichernev:
This is a duplicate of #874 (well the other way around, but I already commented on the other one ;-))

Issue #807, opened 2020-01-01
Wrong timezone used for implicit year/month/day


murrayju:
This is probably only an issue around new year's day. If I parse a date like the following:
```
moment.tz('December 31 5:00PM', 'MMMM D h:mmA', 'America/Denver')
```

My expectation is that it will implicitly use the current year in the `America/Denver` timezone (which is currently 2019). Instead, it seems to be assuming the year that it is in GMT, resulting in the date being parsed as `2021-01-01T00:00:00.000Z` which is a year in the future, instead of today.

murrayju:
Month and day have the same problem. I'm intending to use this to parse times (e.g. `5:00PM`) that implicitly have "today's" date in the given time zone.

Fortunately,
```
moment.tz('America/Denver').format('YYYY-MM-DD')
```
gives the correct/expected result (`2019-12-31`), so I'm able to use this as a dirty workaround:

```
moment.tz(`${moment.tz('America/Denver').format('YYYY-MM-DD')} 5:00PM`, 'YYYY-MM-DD h:mmA', 'America/Denver')
```

ellenaua:
@murrayju how did you get `2021-01-01T00:00:00.000Z`? Did you use `.toISOString()` method?

See my results: 

```
const moment = require('moment-timezone');
const date = moment.tz('December 31 5:00PM', 'MMMM D h:mmA', 'America/Denver')

console.log(date.toISOString()); // 2021-01-01T00:00:00.000Z -- correct, 00:00 UTC
console.log(date.format()); // 2020-12-31T17:00:00-07:00 -- correct, 17:00 in Denver
console.log(date.toString());  // Thu Dec 31 2020 17:00:00 GMT-0700 - correct, 17:00 in Denver
```

ellenaua:
or you used methods like '.getDate()', '.getFullYear()', '.getDate()' which resulted to this:
```
const date = moment.tz('December 31 5:00PM', 'MMMM D h:mmA', 'America/Denver').toDate()
console.log(date.getFullYear()) // 2021
console.log(date.getMonth()) // 0
console.log(date.getDate()) // 1
```

murrayju:
I used `toISOString()`. The point is that I got the value that you currently say is correct, when the local time was last year (2019). The point of the code is to interpret a given time with today's date in a given time zone. It usually works, but there is a window of time on new year's eve that incorrectly gives a date one year in the future instead.

This is a bug report, not a question.

ellenaua:
[...]

PR #1134, opened 2025-06-26
Fix: preserve date when parsing time strings with default timezone

### LINKED ISSUE #874: defaulting to current date-time uses UTC (ignores timezone)
Moment created from formatting time sets the date as the UTC date rather than the date of the default timezone
EX: 

`m = require('moment-timezone')`
`_ = m.tz.setDefault("Asia/Kolkata")`
`console.log(m())`
`console.log(m("T22:00","THH:mm"))`

Output:
moment.parseZone("2020-07-20T03:33:18.943+05:30")
moment.parseZone("2020-07-19T22:00:00.000+05:30")

Sesuraj-git:
fix: preserve local date when parsing time strings with default timezone

Fixes #874, #867, #807

Previously, parsing time-only strings like "T22:00" in a context with
tz.setDefault("Asia/Kolkata") would incorrectly use the UTC date
and shift the local day. This patch uses the original tz parsing
to ensure the correct local date and time are preserved.

Also includes a test case to confirm behavior.

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -588,7 +588,7 @@
 		moment.tz namespace
 	************************************/
 	    var originalTzFn = moment.tz;
-  function tz(input) {
+  	function tz(input) {
         var args = Array.prototype.slice.call(arguments, 0, -1),
             name = arguments[arguments.length - 1],
             zone = getZone(name),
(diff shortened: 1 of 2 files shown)


gilmoreorless:
Thanks for your contribution. Unfortunately, this PR only focuses on string parsing and doesn't fix the root cause (creating instances with only time-based components). Additionally, when I run the new test locally, it fails.

PR #1149, opened 2026-07-16
fix: use correct default day when needed

### LINKED ISSUE #1128: Unexpected date when using moment.tz() with an object with time fields
### Environment

For bug reports, please run the following code in your environment and include the output:

```js
console.log({
  moment: moment.version,
  momentTz: moment.tz.version,
  date: (new Date()).toString(),
  intlZone: new Intl.DateTimeFormat().resolvedOptions().timeZone,
});
```
Output:
```js
{
  moment: '2.30.1',
  momentTz: '0.5.47',
  date: 'Mon Mar 10 2025 23:18:43 GMT+0000 (Coordinated Universal Time)',
  intlZone: 'UTC'
}
```

>How are you loading Moment Timezone? (Script tags, Node.js require(), ES Module imports, etc.)

```js
const moment = require('moment-timezone');
```

### Issue description

Moment returns an unexpected date when using `moment.tz()` with an object with time fields.

Current local timezone: `UTC`  
Current local datetime: `2025-01-02T01:00:00Z`

```js
const timeObject = { hour: 1, minutes: 23, seconds: 56 };
moment.tz(timeObject, 'America/New_York').format();
```

Expected result: `2025-01-01T01:23:56-05:00`  
Actual result: `2025-01-02T01:23:56-05:00`

The object passed to the constructor only has time fields, not date fields.

The [documentation](https://momentjs.com/docs/#/parsing/object/) for Moment parsing with an object says:

>Omitted units default to 0 or the current date, month, and year.

As a result, I would expect the output to use the current date in the specified timezone (`America/New_York`), which is `2025-01-01`, resulting in an output of: `2025-01-01T01:23:56-05:00`.

However, it appears that the constructor is using the current local date (`2025-01-02` in `UTC`), not the date in the specified timezone, which results in an output of: `2025-01-02T01:23:56-05:00`.

Is this the expected behavior?

I can accomplish what I want by doing the following:

```js
const timeObject = { hour: 1, minutes: 23, seconds: 56 };
moment.tz('America/New_York').set(timeObject).format();
```

However, I would assume they would work the same way.

ddelrio1986:
Use the correct default day for the current zone instead of the one for UTC. Closes #1128.

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -588,6 +588,26 @@
 		moment.tz namespace
 	************************************/
 
+	// Re-parse args with moment.now() shifted to the zone's current wall
+	// time, so that any omitted date fields default to "today" in that
+	// zone instead of "today" in UTC. If the input fully specifies its
+	// date fields, this has no effect since the shifted "now" is unused.
+	function todayIn (zone, args) {
+		var realNow  = moment.now(),
+			zoneNow  = realNow - (zone.utcOffset(realNow) * 60000),
+			savedNow = moment.now,
+			out;
+
+		moment.now = function () { return zoneNow; };
+		try {
+			out = moment.utc.apply(null, args);
+		} finally {
+			moment.now = savedNow;
+		}
+
+		return out;
+	}
+
 	function tz (input) {
 		var args = Array.prototype.slice.call(arguments, 0, -1),
 			name = arguments[arguments.length - 1],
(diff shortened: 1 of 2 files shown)


mattjohnsonpint:
Closing in favor of #1156, which is a more comprehensive fix.  But thank you for your contribution anyway, we appreciate it!

PR #1150, opened 2026-07-16
fix: ensure objects are equal when dst on midnight

### LINKED ISSUE #826: Clone creates unequal object when DST starts on midnight
TL;DR `clone` works incorrectly for an object pointing at the midnight DST start

There are several timezones that have DST start/end on midnight, e.g. Havana turns the clock 1 hour forward on Mar 8 00:00, which makes this time technically non-existent. When moving to this date with `add()`,  moment is smart enough to understand it should be Mar 8 start. But if you clone the object, this info is lost and resulting object is not equal to it's source.
```javascript
m = moment.tz('2020-03-01', 'America/Havana')
m.add(1, 'week')
console.log(m.format('YYYY-MM-DD')) // 2020-03-08
console.log(m.clone().format('YYYY-MM-DD')) // 2020-03-07
```

ddelrio1986:
This should ensure that the objects have equal values when a moment sits exactly on a Havana-style DST-shift. Closes #826.

### DIFF
--- a/moment-timezone.js
+++ b/moment-timezone.js
@@ -698,11 +698,28 @@
 		};
 	}
 
+	function cloneWrap (old) {
+		return function () {
+			var m = old.apply(this, arguments);
+			if (this._z) {
+				// The Moment constructor re-derives the offset from
+				// scratch via moment.updateOffset, which can disagree
+				// with the source moment when it sits exactly on a DST
+				// transition boundary. Force the clone to match instead.
+				m._d = new Date(this._d.getTime());
+				m._offset = this._offset;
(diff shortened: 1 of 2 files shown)


gilmoreorless:
Thanks for your contribution. Unfortunately, this doesn't actually fix #826 correctly. Your test of `t.equal(original, "2020-03-08 00:00:00 -04:00", "The original moment should land on the DST start")` is asserting the wrong behaviour - that exact time doesn't exist (it should be `01:00:00`).

Closing this in favour of #1151, which fixed several other cases as well.

